PR #2, opened 2015-01-29
Add images for links that have thumbnails

MaazAli:
Renders an image for links that have thumbnails. It borrows it's code from the Instagram renderer since that essentially does the same thing.


gdub22:
So this is a change to the compiler, which is a separate repo.  Right now the compiler is manually updated in this repo.  Looking into using npm to pull in the code...

MaazAli:
Oops, okay so I should make these changes in that repo and send in a pull request there?

gdub22:
@MaazAli hold off for now.  In looking at that repo I had some unpushed commits that remove all the embed rendering stuff from there.  I believe I refactored it out of there with the idea to move it here because it's more presentation than data.

gdub22:
This has been added another way

PR #9, opened 2015-06-30
Use broccoli, add tests in ci

bantic:
Uses broccoli-multi-builder, runs tests with `npm test` or `http://localhost:4200/tests` and also on CI.

for @mixonic


bantic:
depends on `0.3.1` of content-kit-compiler (see https://github.com/bustlelabs/content-kit-compiler/pull/4)

PR #12, opened 2015-07-02
update demo, add compat layer for win/doc, compile LESS

bantic:
Makes the demo available at http://localhost:4200/demo
I wasn't able to linkify or boldify/italicize text — not sure if that's something I broke in this.


bantic:
@mixonic for you

PR #14, opened 2015-07-09
Drop compiler dependency

mixonic:
Better organize parsers, renderers, etc removing the need for a compiler package. Drop `win` and `doc` as this will never run server-side. Create a runtime directory should should shortly be removed in favor of https://github.com/bustlelabs/mobiledoc-dom-renderer. Remove the concept of a compiler from the editor. Drop many exports. Start converting internals to `post` instead of `model`.


bantic:
sweet. much window, very doc.

PR #15, opened 2015-07-09
Only reference runtime in demo

mixonic:


broccoli/demo.js
```diff
@@ -2,16 +2,25 @@
...
+    funnel(demoDir, {
+      include: [
+        '**/*.css',
+        '**/*.js',
+        'favicon.ico',
+        '**/*.html'
+      ],
+      destDir: '/demo'
+    }),
+    funnel('./node_modules/mobiledoc-dom-renderer/dist/global/', {
```
bantic: I believe it's safest to use path and `require.resolve` here — in npm 3, for instance, deduping may mean we can't guarantee the filepath to the node module for this.

```
// this will be the directory to the file that the renderer's package.json `main` points at
var rendererDir = require('path').dirname( require.resolve('mobiledoc-dom-renderer') );
funnel(rendererDir, {
  include: ['relative/path/from/main/to/mobile-doc-renderer-global'],
  ...
});
```

bantic:
:+1: I would change the way we resolve the path to the file, although now I realize that since we aren't publishing to cjs the [main](https://github.com/bustlelabs/mobiledoc-dom-renderer/blob/master/package.json#L5) entry is invalid anyway. I'll change it to point at the global build.

bantic:
It's still a little weird for main to point to a non-commonjs file, so I could go either way on this one.

PR #18, opened 2015-07-13
Event listeners

bantic:
Adds an `EventListener` mixin that is mixed into the base View class, the Editor, and the ToolbarButton. It adds an `addEventListener` method that keeps track of added listeners so that the mixed-in `removeAllEventListeners` method can be called later.

### DIFF
--- a/src/js/utils/event-listener.js
+++ b/src/js/utils/event-listener.js
@@ -1,0 +1,14 @@
+export default class EventListenerMixin {
+  addEventListener(context, eventName, listener) {
+    if (!this._eventListeners) { this._eventListeners = []; }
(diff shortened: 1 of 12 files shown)


src/js/utils/event-listener.js
```diff
@@ -0,0 +1,14 @@
+export default class EventListenerMixin {
```
mixonic: Why not keep mixins as pain objects? Used this way, the `prototype` is basically just a bag for the mixin properties. An object would be sufficient.

mixonic:
Very cool, but needs a rebase.

We should really nuke utils.

PR #19, opened 2015-07-13
compatibility fixes for IE10

bantic:
IE10 doesn't support `dataset` or `DOMNode.remove()`. This PR papers over those cases.
The tests do no yet pass on IE10, and the editor demo also doesn't work. Largely due to the fact that IE10 does not fire the `input` event on contenteditables. (See compatibility notes at [MDN doc](file:///Users/coryforsyth/Library/Application%20Support/Dash/DocSets/JavaScript/JavaScript.docset/Contents/Resources/Documents/developer.mozilla.org/en-US/docs/Web/Events/input.html)).


mixonic:
derp- if we keep needing compat items we should consider bringing in an existing lib.

PR #23, opened 2015-07-19
Post, Section, Markers

bantic:
WIP to use `Post`, `Section`, `Marker`, `Markup` models to build the internal `post` model that the editor has.

I started this as a WIP to see what it would take to be able to semantically handle section-breaking actions (like hitting `<enter>` or clicking the "h2" on the toolbar) and section-joining actions (highlighting text across sections and deleting, putting the cursor at the beginning of a section and deleting). I started with the `Marker` concept representing text with possible markups applied to it and thinking of the methods a `Marker` would need in order to handle the cases mentioned above. I rewrote the post/section parser to use the marker/markup construct internally as well. This fixes the problem of ignorable (ie `<span>`) elements causing the parser to generate separate markers for the same string of text.

A `Post` hasMany `sections`, a `Section` hasMany `markers`, a `Marker` has a tagName and many `markups`.

Adds a `PostParser` (from DOM to Post) and its `SectionParser`. The `PostParser` is simpler than the existing dom parser, partly because it makes the simplifying assumption that if it is parsing dom into a post, only the top-level element's children are potentially valid sections. It will ignore invalid (from its perspective) dom like a nested section tag inside another (`<p>first section<p>inner section</p> continue first section</p>`).

In order to be usable, these need to happen:
  * [ ] Enhance SectionParser to parse Image sections
  * [ ] Change Mobiledoc renderer to consume this format (the major change is that markers don't have `open` and `close` properties on them anymore)
  * [ ] The post-builder is probably not necessary, because all the primitives exist and have their own constructors (`new Section()`, `new Marker()`, `new Markup()`).

The goal of this was to eventually be able to add event listeners for section-breaking (`<enter>` key) and section-joining (`delete` key) actions that the user might take and apply those directly to the internal `post` model rather than the current heuristic of attempting to detect which sections might have been affected or removed.

### DIFF
--- a/src/js/models/section.js
+++ b/src/js/models/section.js
@@ -1,12 +1,15 @@
-export const DEFAULT_TYPE = 'p';
-export const SECTION_TYPES = [
+export const DEFAULT_TAG_NAME = 'p';
+export const SECTION_TAG_NAMES = [
   'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'div'
 ];
+const SECTION_TYPE = 'section';
 
 const Section = class Section {
-  constructor(type=DEFAULT_TYPE, markers=[]) {
+  constructor(tagName=DEFAULT_TAG_NAME, markers=[]) {
     this.markers = [];
-    this.type = type;
+    this.tagName = tagName;
+    this.type = SECTION_TYPE;
+
     markers.forEach(m => this.appendMarker(m));
   }
 
@@ -36,8 +39,8 @@ const Section = class Section {
     right.push(rightMiddle);
 
     return [
-      new Section(this.type, left),
-      new Section(this.type, right)
+      new Section(this.tagName, left),
+      new Section(this.tagName, right)
     ];
   }
 
--- a/src/js/models/section.js
+++ b/src/js/models/section.js
@@ -1,0 +1,66 @@
+export const DEFAULT_TYPE = 'p';
+export const SECTION_TYPES = [
+  'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'div'
+];
+
+const Section = class Section {
+  constructor(type=DEFAULT_TYPE, markers=[]) {
+    this.markers = [];
+    this.type = type;
+    markers.forEach(m => this.appendMarker(m));
+  }
+
+  appendMarker(marker) {
+    this.markers.push(marker);
+  }
+
+  /**
+   * @return {Array} 2 new sections
+   */
+  split(offset) {
+    let left = [], right = [], middle;
+
+    middle = this.markerContaining(offset);
+    const middleIndex = this.markers.indexOf(middle);
+
+    for (let i=0; i<this.markers.length; i++) {
+      if (i < middleIndex) { left.push(this.markers[i]); }
+      if (i > middleIndex) { right.push(this.markers[i]); }
+    }
+
+    let leftLength = left.reduce((prev, cur) => prev + cur.length, 0);
+    let middleOffset = offset - leftLength;
+
+    let [leftMiddle, rightMiddle] = middle.split(middleOffset);
+    left.push(leftMiddle);
+    right.push(rightMiddle);
+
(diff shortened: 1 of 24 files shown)


mixonic:
A few items, going to try and make some progress today

> A Post hasMany sections, a Section hasMany markers, a Marker has a tagName and many markups.

*some* sections have markers, yes. Specifically, markup sections. I'm starting to think we should not ship  image sections, but just implement them as cards :-p Unsure if this is an over abstraction.

> Adds a PostParser (from DOM to Post) and its SectionParser

Totally agree with the sentiment. The names `dom-parser` and `mobiledoc-parser` drew attention to what was being parsed. I'd prefer to call these new versions `post-dom-parser` `section-dom-parser`

src/js/parsers/section.js
```diff
@@ -0,0 +1,111 @@
...
+  isSectionElement(element) {
+    return element.nodeType === ELEMENT_NODE &&
+      SECTION_TAG_NAMES.indexOf(element.tagName.toLowerCase()) !== -1;
+  },
+
+  markupFromElement(element) {
+    const tagName = element.tagName.toLowerCase();
+    if (MARKUP_TAG_NAMES.indexOf(tagName) === -1) { return null; }
+
+    return new Markup(tagName, getAttributes(element));
```
mixonic: This should be a whitelist IMO, not just all attributes.

src/js/renderers/mobiledoc.js
```diff
@@ -5,6 +5,10 @@ let visitor = {
     opcodes.push(['openPost']);
     visitArray(visitor, node.sections, opcodes);
   },
+  section(node, opcodes) {
```
mixonic: Should be a `markupSection` type

src/js/parsers/post.js
```diff
@@ -0,0 +1,19 @@
+import Post from 'content-kit-editor/models/post';
+import SectionParser from 'content-kit-editor/parsers/section';
+import { forEach } from 'content-kit-editor/utils/array-utils';
+
+export default {
+  parse(element) {
+    const post = new Post();
+
+    forEach(element.childNodes, child => {
+      post.appendSection(SectionParser.parse(child));
```
mixonic: Hm, right so this requires a pre-processing step to the DOM before it can accept input from paste or upon page-load.

src/js/models/section.js
```diff
@@ -0,0 +1,69 @@
+export const DEFAULT_TAG_NAME = 'p';
+export const SECTION_TAG_NAMES = [
+  'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'div'
```
mixonic: Seems like this should be the same list as https://github.com/bustlelabs/content-kit-editor/blob/summertime/src/js/parsers/dom.js#L7

src/js/parsers/section.js
mixonic: Like here https://github.com/bustlelabs/content-kit-editor/blob/summertime/src/js/parsers/dom.js#L9

mixonic:
I've moved the existing render logic over to the new models. I'm unsure how we can use the parsing parts themselves though- there are several tricky things that need to be done to handle incoming payloads that may not be sane.

I'm also unsure why we should be `toLowerCase()` on tag names. The capitalized tag names were doing no harm- is it only for ascetics or is there a problem with the caps that they must be normalized?

Attribute list still needs to be sanitized.

I've removed some logic that disallowed the same tag name to be added twice at the marker level. This seems like it should be a parsing concern, is there a specific use-case you have in mind? Multiple tags of a span or em type are in theory harmless.

mixonic:
As discussed with @bantic, we are going to merge this into summertime now. Further changes will continue to diverge this codebase further and make code review increasingly difficult.

At some later point (probably after Cory's next patch lands), we should get rid of any extraneous parts of the lib. Specifically I'm thinking the two new parsers added here for trusted DOM, which are not actually in any used code outside of tests.

PR #29, opened 2015-07-29
Pass version number and sections payload on mobiledocs

mixonic:
Renderers must also be updated before this can be merged.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -289,99 +289,140 @@ function attemptEditorReboot(editor, textarea) {
   }
 }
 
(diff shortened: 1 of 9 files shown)


demo/demo.js
```diff
@@ -289,99 +289,140 @@ function attemptEditorReboot(editor, textarea) {
   }
 }
 
+var MOBILEDOC_VERSION = "0.1";
```
bantic: We should probably semver this as "0.1.0" ?

mixonic: Hm. I agree.

PR #31, opened 2015-08-03
WIP: Fix embed-intent

mixonic:
fix the embed intent.

TODO: 

* [ ] Test


mixonic:
The embed intent can definitely be optimized- It can fire based on `input` or `rerender` on the editor instead of `keyup`.

PR #32, opened 2015-08-03
Refactor editor to delegate selection methods to `Cursor`

bantic:
* remove unused methods from cursor and editor
 * add `ReversibleToolbarButton` to represent a button that can be `exec`'d and `unexec`'d. Some duplication with the code for buttons that already exist, but this moves closer to being able to clean up all buttons
 * Change the `TextFormatToolbar` to be instantiated with buttons in addition to commands. Eventually only buttons should be added to it.
 * Change all the block format commands (heading, subheading, 
 * Several tests surrounding applying/unapplying block format headings and ensuring that their buttons stay active/inactive as necessary
 * small change: heading button now says "h2" and subheading button says "h3", because those are the tags that they apply
 * add `normalizeTagName` method in dom utils, use it for methods interacting with markup sections.

>>  * `Post.insertSectionAfter` can take a null reference section, like the DOM method `insertAfter`

I was wrong about that ^^. For one thing the DOM node method is `insertBefore` and it's unclear from the docs whether it can always deal with a null reference node. I may change this back at some point.

I believe this is ready to go, now. Next step is to add code to split markers at the text selection, which unblocks the ability to apply markup commands (bold, italic, etc). I'll address that in a separate PR.

cc @mixonic

### DIFF
--- a/src/js/commands/format-block.js
+++ b/src/js/commands/format-block.js
@@ -1,8 +1,21 @@
 import TextFormatCommand from './text-format';
+import {
+  any
+} from '../utils/array-utils';
 
 class FormatBlockCommand extends TextFormatCommand {
-  constructor(options={}) {
+  constructor(editor, options={}) {
     super(options);
+    this.editor = editor;
+  }
+
+  isActive() {
+    const editor = this.editor;
+    const activeSections = editor.activeSections;
+
+    return any(activeSections, section => {
+      return any(this.mappedTags, t => section.tagName === t);
+    });
   }
 
   exec() {
@@ -15,8 +28,6 @@ class FormatBlockCommand extends TextFormatCommand {
     });
 
     editor.rerender();
-    editor.trigger('update'); // FIXME -- should be handled by editor
-
     editor.selectSections(activeSections);
   }
 
@@ -29,8 +40,6 @@ class FormatBlockCommand extends TextFormatCommand {
     });
 
     editor.rerender();
-    editor.trigger('update'); // FIXME -- should be handled by editor
-
     editor.selectSections(activeSections);
   }
 }
--- a/src/js/commands/format-block.js
+++ b/src/js/commands/format-block.js
@@ -1,36 +1,38 @@
 import TextFormatCommand from './text-format';
-import { getSelectionBlockElement, selectNode } from '../utils/selection-utils';
-import { inherit } from 'content-kit-utils';
 
-function FormatBlockCommand(options) {
-  options = options || {};
-  options.action = 'formatBlock';
-  TextFormatCommand.call(this, options);
-}
-inherit(FormatBlockCommand, TextFormatCommand);
-
-FormatBlockCommand.prototype.exec = function() {
-  var tag = this.tag;
-  // Brackets neccessary for certain browsers
-  var value =  '<' + tag + '>';
-  var blockElement = getSelectionBlockElement();
-  // Allow block commands to be toggled back to a text block
-  if(tag === blockElement.tagName.toLowerCase()) {
-    throw new Error('Unimplemented: Type.BOLD.paragraph must be replaced');
-    /*
-    value = Type.PARAGRAPH.tag;
-    */
-  } else {
-    // Flattens the selection before applying the block format.
-    // Otherwise, undesirable nested blocks can occur.
-    // TODO: would love to be able to remove this
-    var flatNode = document.createTextNode(blockElement.textContent);
-    blockElement.parentNode.insertBefore(flatNode, blockElement);
-    blockElement.parentNode.removeChild(blockElement);
-    selectNode(flatNode);
+class FormatBlockCommand extends TextFormatCommand {
+  constructor(options={}) {
+    super(options);
   }
 
-  FormatBlockCommand._super.prototype.exec.call(this, value);
-};
+  exec() {
+    const editor = this.editor;
+    const activeSections = editor.activeSections;
+
+    activeSections.forEach(s => {
+      editor.resetSectionMarkers(s);
+      editor.setSectionTagName(s, this.tag);
+    });
+
+    editor.rerender();
+    editor.trigger('update'); // FIXME -- should be handled by editor
+
+    editor.selectSections(activeSections);
+  }
+
+  unexec() {
+    const editor = this.editor;
+    const activeSections = editor.activeSections;
+
+    activeSections.forEach(s => {
+      editor.resetSectionTagName(s);
+    });
+
+    editor.rerender();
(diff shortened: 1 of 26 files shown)


src/js/commands/format-block.js
```diff
@@ -1,36 +1,24 @@
...
 
-  FormatBlockCommand._super.prototype.exec.call(this, value);
-};
+  exec() {
+    const editor = this.editor;
+    const activeSections = editor.activeSections;
+
+    activeSections.forEach(s => {
+      editor.resetSectionMarkers(s);
+      editor.setSectionTagName(s, this.tag);
```
mixonic: curious why these are methods on the editor instead of on the sections itself

src/js/editor/editor.js
```diff
@@ -535,21 +510,23 @@ class Editor {
...
-        if (section.renderNode) {
-          section.renderNode.scheduleForRemoval();
-        } else {
-          throw new Error('All sections are expected to have a renderNode');
-        }
+        deletedSections.push(section);
       }
-    }
+    });
+    forEach(deletedSections, (s) => s.renderNode.scheduleForRemoval());
```
mixonic: Why pull this into an array to be traversed a second time?

src/js/editor/editor.js
```diff
@@ -648,7 +622,7 @@ class Editor {
 
   insertSectionAtCursor(newSection) {
     let newRenderNode = this._renderTree.buildRenderNode(newSection);
-    let renderNodes = this.getRenderNodesWithCursor();
+    let renderNodes = this.cursor.activeSections.map(s => s.renderNode);
```
mixonic: `getRenderNodesWithCursor` returned render nodes for the markers- `activeSections` will return the sections?

Besides the confusion about that change, I'm unsure if it makes more sense to be talking about "what markers are selected" vs "what render nodes are selected". It is understandable that you can always find one from the other, but my leaning has been to make the render node the object interacted with by the editor as much as possible. This turns it around a little to be focused on the dom-ignorant post nodes.

src/js/parsers/post.js
```diff
@@ -52,7 +52,7 @@ export default {
 
     // walk up from the textNode until the rootNode, converting each
     // parentNode into a markup
-    function collectMarkups(textNode, rootNode) {
+    const  collectMarkups = (textNode, rootNode) =>{
```
mixonic: using `const foo = () => {` is pretty harmless, we should wander and discuss if it should be used for basically any function though. I'm unclear why some are migrated to the new style and others are not, or why the new style is superior.

src/js/utils/dom-utils.js
```diff
@@ -112,6 +111,11 @@ function getAttributesArray(element) {
   return result;
 }
 
+function addClassName(element, className) {
+  // FIXME-IE IE10+
```
mixonic: seems ok.

tests/unit/editor/editor-test.js
```diff
@@ -35,7 +35,16 @@ test('creating an editor without a class name adds appropriate class', (assert)
...
   var editor = new Editor(document.getElementById('editor1'));
-  assert.equal(editor.element.className, 'ck-editor');
+  assert.equal(editor.element.className, EDITOR_ELEMENT_CLASS_NAME);
+});
+
+test('creating an editor adds EDITOR_ELEMENT_CLASS_NAME if not there', (assert) => {
+  editorElement.className = 'abc def';
+
+  var editor = new Editor(document.getElementById('editor1'));
+  const hasClass = (className) => editor.element.className.indexOf(className) !== -1;
```
mixonic: Also:

```
editor.element.classList.contains(className)
```

src/js/editor/editor.js
bantic: The idea was to use an expressive variable `deletedSections` to describe why the sections with render nodes were being scheduled for removal. I am not sure if the perf impact of a second loop through the sections is significant. If it seems so I can back this out and lean on the comment (`// remove deleted nodes`) above. I didn't see that comment when refactoring this before.

src/js/commands/format-block.js
bantic: I've been thinking of the editor model as the keeper of the public api, and these commands as plugin-style consumers of that public api. It has made me reluctant to have them call methods on the constituent objects directly. Delegating methods through the editor allows it to also take care of its own render-related bookkeeping. I can see the utility of considering methods like `section#setTagName` (or similar) as part of the public API, but it seems like they may cause the sections and markers to end up more tightly coupled to the render tree as a result.

src/js/utils/dom-utils.js
bantic: yeah — I'm adding `FIXME-IE` as breadcrumbs for an IE audit at some point in the future.

tests/unit/editor/editor-test.js
bantic: +1 I'll change it

src/js/parsers/post.js
bantic: I think when creating functions to use in blocks (i.e. helper functions in other functions), it makes some sense to declare them in a variable-ish way like this. Mostly, though, I think I just enjoy the new syntax and I convert named function expressions over to this if I stare at them too long :P

src/js/editor/editor.js
bantic: Yeah, perhaps we can start to rectify these into a unified approach soon. I have been thinking of the `post` as the single source of truth, partly because historically renderNodes for a given postNode did not always exist until, so they require some mental bookkeeping to keep track when the last render was, etc. The object model of posts->sections->markers->markups helps me to keep everything conceptually straight as well — some edit actions apply only to one instead of the other (ie block-format commands like h1, h2 apply to sections, text format commands like b, i apply to markers), and I like that the code makes that distinction clear.

bantic: I think we are close to capturing enough of the edit actions to be able to ensure that renderNodes and postNodes stay in sync, and at that point it may make more sense to move more of the actions through renderNodes directly. The extra boilerplate around render nodes (build render node, attach it, mark it clean/dirty) I think adds a little friction to using them as freely, but that can be easily changed.

src/js/editor/editor.js
mixonic: I'll merge, but yeah traversing a second time seems unnecessary to me.

src/js/parsers/post.js
mixonic: FWIW I find this pattern (creating functions inside a function scope) to be very un-friendly to read, and have foot-gun-esq problems of scope (what variables are leaking in etc). Unless we're generating callbacks, I'm not crazy about using locally scoped function for helpers.

PR #33, opened 2015-08-04
Post images to be cards

mixonic:
TODO:

* Add a default image card implementation to mobiledoc DOM and HTML renderers


demo/demo.js
```diff
@@ -182,7 +182,8 @@ var ContentKitDemo = exports.ContentKitDemo = {
       'simple-card': simpleCard,
       'edit-card': cardWithEditMode,
       'input-card': cardWithInput,
-      'selfie-card': selfieCard
+      'selfie-card': selfieCard,
+      'image': ContentKit.ImageCard
```
mixonic: need to add this to the list if not present (in the editor constructor)

mixonic: Not in the editor constructor, but should be added to the default renderer (some version should be there)

src/js/commands/image.js
```diff
@@ -1,74 +1,24 @@
...
-          this.editor.trigger('update');
-        } else {
-          this.editor.removeSection(imageSection);
-          new Message().showError(error.message || 'Error uploading image');
-        }
-        this.editor.rerender();
-      }
-    });
+    let {post} = this.editor;
+    let sections = this.editor.cursor.activeSections;
```
bantic: should just be `this.editor.activeSections`

src/js/commands/image.js
```diff
@@ -1,74 +1,24 @@
...
+    let lastSection = sections[sections.length - 1];
+    let section = this.builder.generateCardSection('image');
+    post.insertSectionAfter(section, lastSection);
+    sections.forEach(section => section.renderNode.scheduleForRemoval());
```
bantic: why does it remove all the sections before this card?

tests/acceptance/embed-intent-test.js
```diff
@@ -43,3 +61,25 @@ Helpers.skipInPhantom('typing inserts section', (assert) => {
...
+  Helpers.dom.triggerEvent(editorElement.childNodes[1].firstChild, 'click');
+
+  let done = assert.async();
+  setTimeout(() => { // delay due to internal async
+    assert.ok($('.ck-embed-intent').is(':visible'), 'embed intent appears');
+
+    Helpers.dom.triggerEvent($('.ck-embed-intent-btn')[0], 'click');
+    Helpers.dom.triggerEvent($('button[title=image]')[0], 'click');
+
+    assert.ok($('.ck-embed-intent').length === 0, 'embed intent is hidden');
```
bantic: there is `assert.hasElement` and `assert.hasNoElement`

bantic:
:+1: looks good to me. I had a few questions but go ahead and merge at will

src/js/commands/image.js
mixonic: the selected sections, or just the section the cursor is on, should be replaced. The common scenario is that the intent tools are showing on a blank line and that blank line is replaced with the image.

mixonic:
Upstream mobiledoc renderers now have support for default image cards.

demo/demo.js
mixonic: upstream lib now has a default.

PR #38, opened 2015-08-05
Inline commands

bantic:
* Changes `bold` and `italic` buttons to apply changes semantically (to the AT directly via `editor.applyMarkupToSelection` and `editor.removeMarkupFromSelection`)

  * Changes the tag name for bold to be `<strong>`. There may be a case here that the tag it should use is `<b>` because the intent may not be semantic (i.e. the author really intended to indicate a stylistic change rather than express emphasis). I could go either way, but I think italic and bold should either both be structural/semantic tags (em/strong) or both be style tags (i/b). This is easy to change back.

  * Adds static method `Markup.create` and changes all usage of `new Markup` to it so that we can ensure we get singleton markups for single-tag markup.

  * Uses `normalizeTagName` everywhere instead of sprinkled `toLowerCase` calls – along with using `Markup.create` this helps to ensure that we are always dealing with the same markup tag (so that the "B" button in the toolbar is always correctly active/inactive).

  * Fixed a bug that occurred when selecting multiple sections due to `Section#nextSibling` choking on `this.post` — now whenever a post adds a section (via `appendSection` and other methods, we set the section's post, just like we do when adding markers to a section). There's a test covering this behavior in the editor-selections-test file.

cc @mixonic


mixonic:
:-D have not reviewed code yet, but we discussed replacing `Markup.create` with a builder.

```js
// is editor constructor
this.builder = new PostNodeBuilder();

// then you can
this.builder.createMarkup('B')
this.builder.createMarker('some text', markups)

// parsers should have the builder passed in, and objects
// returned from the createFoo methods should have the
// builder set on them as a property before they are returned
```

mixonic:
`normalizeTagName` +1

ErisDS:
I don't know whether this is intended to be fixed here or no:

If you load a Mobiledoc with a marker (e.g. the currently commented out `mobileDocWithMarker` from the demo) the matching toolbar item isn't blue/highlighted i.e. isActive isn't correct. Please ignore me if it's not meant to be working yet, I'm just poking things and thought I'd mention it.

bantic:
@ErisDS Thanks, that is a good catch. That should be *mostly* fixed by this PR. The likely difference is that the `mobileDocWithMarker` uses "B" but in this PR I changed the bold button to use the "strong" tag. The "B" button on the toolbar probably needs to be smart enough to recognize "b" or "strong" tagged markers.

ErisDS:
Hi, I've been looking at it some more and literally just come to the same conclusion :+1:

bantic:
Added #43 to track the bold toolbar button activating on either 'B' or 'STRONG'.

bantic:
@mixonic I'm going to add on a commit to this PR to use the builder everywhere, as discussed above

bantic:
@mixonic 0c20232 removes `post-builder` and uses a 1-instance-per-editor-instance `PostNodeBuilder`.

This PR is good to go, now.

src/js/commands/italic.js
```diff
@@ -1,14 +1,25 @@
...
-    button: '<i class="ck-icon-italic"></i>'
-  });
+export default class ItalicCommand extends TextFormatCommand {
+  constructor(editor) {
+    super({
+      name: 'italic',
+      button: '<i class="ck-icon-italic"></i>'
+    });
+    this.editor = editor;
+    this.markup = Markup.create('em');
```
mixonic: so this should be through the builder

src/js/models/markup-section.js
```diff
@@ -129,4 +152,11 @@ export default class Section {
       return this.markers[leftInclusive ? i : i-1];
     }
   }
+
+  get nextSibling() {
+    const index = this.post.sections.indexOf(this);
```
mixonic: funky highlighting

src/js/models/markup.js
```diff
@@ -22,8 +27,26 @@ export default class Markup {
...
   }
 
+  // Use `create` to make a new markup so that we can use singletons in the
+  // markupMap
+  static create(tagName, attributes=[]) {
+    tagName = normalizeTagName(tagName);
+
+    if (attributes.length === 0) {
+      if (!markupMap[tagName]) {
+        markupMap[tagName] = new Markup(tagName);
```
mixonic: should use builder

mixonic: Oh right, this whole function should perhaps not exist (builder should be doing this instead).

tests/unit/models/marker-test.js
```diff
@@ -29,22 +29,22 @@ test('a marker can truncated to an offset', (assert) => {
 
 test('a marker can have a markup applied to it', (assert) => {
   const m1 = new Marker('hi there!');
-  m1.addMarkup(new Markup('b'));
+  m1.addMarkup(Markup.create('b'));
```
mixonic: Unsure these tests should exist like this- perhaps they should go though the builder?

mixonic:
This is an amazing cleanup

PR #41, opened 2015-08-05
Fixes mobiledoc sample & removes jQuery from usage

ErisDS:
Ok, I know, this is a totally lame attempt at a first commit :grin: 

I noticed there was a missing closing bracket in the 'usage' code example in the README.md.

Whilst I was there, I also swapped out the jQuery snippet to one that doesn't require jQuery. This is totally a personal preference thing so am happy to revert that if you prefer to keep it as it was.

- simpleMobiledoc sample was missing a closing bracket
- switched from jQuery to document.querySelector

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,9 @@ var simpleMobiledoc = {
     [1, "p", [
       [[], 0, "Welcome to Content-Kit"]
     ]]
-  ]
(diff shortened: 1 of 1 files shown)


mixonic:
:-D Thanks @ErisDS!

PR #44, opened 2015-08-06
37 selection deletion

bantic:
*do not merge*
After #38 is merged I'll rebase this.

This adds `post#cutMarkers(markers)` and `editor#deleteSelection`.

To do:
  * [ ] handle newlines when there is a selection (delete the selection first then apply the newline action)
  * [ ] handle keystrokes when there is a selection (contenteditable almost does this for us, but it inserts irrelevant spans to try to maintain visual styling — this should be handled semantically as well)


bantic:
closing to reopen with `master` as the merge target

PR #46, opened 2015-08-06
37 selection deletion

bantic:
This adds `post#cutMarkers(markers)` and `editor#deleteSelection`.

Also adds `Helpers.toolbar` for sharing toolbar-related assertions in tests, and
`Helpers.dom.triggerDelete`, which detects phantom and simulates a delete (by calling `editor.handleDeletion`) — this allows some of the previous skipped-in-phantom tests to run in phantom, now.

Ready for review. I will start on the items below but make a separate PR for them. @mixonic lmk if you have feedback.

To do:
  * [ ] handle newlines when there is a selection (delete the selection first then apply the newline action) (#49)
  * [ ] handle keystrokes when there is a selection (contenteditable almost does this for us, but it inserts irrelevant spans to try to maintain visual styling — this should be handled semantically as well) (#50)

### DIFF
--- a/src/js/models/marker.js
+++ b/src/js/models/marker.js
@@ -6,10 +6,6 @@ import {
 import { detect } from 'content-kit-editor/utils/array-utils';
 
 const Marker = class Marker {
-  static createBlank() {
-    return new Marker('');
-  }
-
   constructor(value='', markups=[]) {
     this.value = value;
     this.markups = [];
--- a/src/js/models/marker.js
+++ b/src/js/models/marker.js
@@ -6,6 +6,10 @@ import {
 import { detect } from 'content-kit-editor/utils/array-utils';
 
 const Marker = class Marker {
+  static createBlank() {
+    return new Marker('');
+  }
+
   constructor(value='', markups=[]) {
     this.value = value;
     this.markups = [];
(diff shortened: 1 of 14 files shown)


src/js/models/post.js
```diff
@@ -19,6 +20,53 @@ export default class Post {
...
+      }
+
+      currentSection = marker.section;
+      currentSection.removeMarker(marker);
+    });
+
+    // add a blank marker to any sections that are now empty
+    changedSections.forEach(section => {
+      if (section.isEmpty()) {
+        section.appendMarker(Marker.createBlank());
```
mixonic: Should go through the builder?

src/js/models/marker.js
```diff
@@ -6,6 +6,10 @@ import {
 import { detect } from 'content-kit-editor/utils/array-utils';
 
 const Marker = class Marker {
+  static createBlank() {
```
mixonic: Should be defined on the builder?

bantic: yep, thanks. I'll change this.

mixonic:
@bantic beside that comment on `createMarker` I think this looks mighty good. Do you want to merge?

mixonic:
Because if you merge before #48 I'll do the rebase for `cutMarkers` instead of making you do it :wink:

bantic:
:) thanks! I can help with rebasing cutMarkers if you need it.

PR #47, opened 2015-08-07
Introduce general implementation of a linked list

mixonic:
Should be widely useful across the codebase, this commit adds the implementation and ports render nodes to the new base class.

### DIFF
--- a/src/js/models/render-node.js
+++ b/src/js/models/render-node.js
@@ -1,75 +1,47 @@
-export default class RenderNode {
+import LinkedItem from "content-kit-editor/utils/linked-item";
+import LinkedList from "content-kit-editor/utils/linked-list";
+
+export default class RenderNode extends LinkedItem {
   constructor(postNode) {
-    this.parentNode = null;
+    super();
+    this.parent = null;
     this.isDirty = true;
     this.isRemoved = false;
     this.postNode = postNode;
-
-    this.firstChild = null;
-    this.lastChild = null;
-    this.nextSibling = null;
-    this.previousSibling = null;
+    this.childNodes = new LinkedList({
+      adoptItem: item => {
+        item.parent = this;
+        item.renderTree = this.renderTree;
+      },
+      freeItem: item => {
+        item.parent = null;
+        item.renderTree = null;
+      }
+    });
   }
   scheduleForRemoval() {
     this.isRemoved = true;
-    if (this.parentNode) {
-      this.parentNode.markDirty();
+    if (this.parent) {
+      this.parent.markDirty();
     }
   }
   markDirty() {
     this.isDirty = true;
-    if (this.parentNode) {
-      this.parentNode.markDirty();
+    if (this.parent) {
+      this.parent.markDirty();
     }
   }
   markClean() {
     this.isDirty = false;
   }
   appendChild(child) {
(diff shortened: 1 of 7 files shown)


src/js/models/render-node.js
```diff
@@ -1,75 +1,47 @@
...
-      this.firstChild = child;
-    }
-    if (this.lastChild) {
-      child.previousSibling = this.lastChild;
-      this.lastChild.nextSibling = child;
-    }
-    this.lastChild = child;
-    child.parentNode = this;
-    child.renderTree = this.renderTree;
+    this.childNodes.append(child);
```
bantic: yes!

src/js/models/render-node.js
```diff
@@ -1,75 +1,47 @@
-export default class RenderNode {
+import LinkedItem from "content-kit-editor/utils/linked-item";
+import LinkedList from "content-kit-editor/utils/linked-list";
+
+export default class RenderNode extends LinkedItem {
```
bantic: I would prefer composition over inheritance here — we have a function in utils called `mixin` that can be used instead. This is fine for now, though.

src/js/utils/linked-list.js
```diff
@@ -0,0 +1,84 @@
+export default class LinkedList {
+  constructor(options) {
```
bantic: we can kill some of the conditionals in the rest of this file if we assign no-op defaults to `adoptItem` and `freeItem`:

```
constructor({adoptItem: () => {}, freeItem: () => {}}
```

And then no need to check `if this.adoptItem` in `insertBefore` or `if this.freeItem` in `remove`

src/js/utils/linked-list.js
```diff
@@ -0,0 +1,84 @@
...
+    this.insertBefore(item, null);
+  }
+  insertAfter(item, prevItem) {
+    let nextItem = null;
+    if (prevItem) {
+      nextItem = prevItem.next;
+    }
+    this.insertBefore(item, nextItem);
+  }
+  insertBefore(item, nextItem) {
```
bantic: I find these arg names mildly confusing — `insertBefore(newItem, oldItem)` makes more sense to me.

I think this can be simplified:
```
newItem.prev = oldItem.prev;
newItem.next = oldItem;

if (oldItem.prev) {
  oldItem.prev.next = newItem;
} else {
  this.head = newItem;
}

oldItem.prev = newItem;
```

bantic: Actually that ^^ is still wrong. I had to pull this and branch off it to make sure I had it right. Here's https://github.com/mixonic/content-kit-editor/pull/1, which merges onto this PR, if you like where those ^^ comments were going.

bantic:
This makes the code so much cleaner. After my comments on the LL impl went off the rails above I took a pass at simplifying the linked list in https://github.com/mixonic/content-kit-editor/pull/1. Back to you to merge in my PR if you want, otherwise :+1: .

src/js/models/render-node.js
mixonic: I think we should make this a decorator at some point.

mixonic:
As discussed, I'm going to merge this and move forward. There were some tricky design decisions here I should have done a better job of explaining when I opened this.

PR #48, opened 2015-08-08
Port posts and markers to linked lists

mixonic:
Continues with work begun with https://github.com/bustlelabs/content-kit-editor/pull/47

* Adds some new helper methods:
  * `forEach` with an iterator and the index passed to a callback
  * `takeRange(startItem, endItem)` if the start is falsy, we presume `head`. If the end is falsy, we presume `tail`.
  * `toArray` which returns an array.
  * `detect` with takes a callback expected to return `true` for a match. Takes a second argument for a starting marker
* Some editing oddness remains, but tests pass and I think that bad behavior is related to pre-existing issues (covered in https://github.com/bustlelabs/content-kit-editor/pull/46 and in future PRs)
* Makes the `childNodes` property on render nodes lazy.
* In some places, we are building arrays that are later passed to a new section or merged onto an existing one. I wonder if some of that could be improved by not building and splicing in an array but simply another linked list. The nodes would still need to be iterated and adopted, but the array would never be needed.

TODO:

* [x] rename `takeRange` to `readRange`, since it does not actually take anything
* [x] ~~add `replace` helper (allows some other replace code to be shared). Should take two arguments, first the item to be replaced and second an array of replacements.~~ Actually implemented a `splice` method. First arg is a marker, second a count of how many markers should be removed including and after that marker, third argument is an array of new markers.
* [x] drop custom append, prepend, replace, insert, remove etc methods in favor of directly interacting with the linked list.
* [x] add `objectAt` to replace `head.next.next.next` junk in tests. We don't have any need for it in the application code itself.

### DIFF
--- a/src/js/models/marker.js
+++ b/src/js/models/marker.js
@@ -7,9 +7,11 @@ import {
   detect,
   difference
 } from 'content-kit-editor/utils/array-utils';
+import LinkedItem from "content-kit-editor/utils/linked-item";
 
-const Marker = class Marker {
+const Marker = class Marker extends LinkedItem {
   constructor(value='', markups=[]) {
+    super();
     this.value = value;
     this.markups = [];
     this.type = MARKER_TYPE;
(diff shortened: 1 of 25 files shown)


src/js/models/marker.js
```diff
@@ -119,32 +121,18 @@ const Marker = class Marker {
...
-      return this.section.markers[index + 1];
-    }
-  }
-
-  get previousSibling() {
-    let index = this.section.markers.indexOf(this);
-    if (index > 0) {
-      return this.section.markers[index - 1];
-    }
-  }
```
bantic: :clap:

bantic:
this is great.

mixonic:
Cool, just the big custom method dropping to do.

mixonic:
:-D :-D :-D :-D :-D :-D :fireworks:

PR #53, opened 2015-08-10
Handle newline at start or end of section

### LINKED ISSUE #39: Pressing enter at the start of a line/section has unexpected results
Placing the cursor at the start of a line and pressing `enter` causes the line to move down in 0.2.1. In 0.2.2 it at first appears to do nothing.

If, after having pressed `enter` at the beginning of a line, you move the cursor to somewhere else on the same line and also press `enter`, the content of the line is duplicated before the new line is added.

How I reproduced:

- on master (58ffbc8) using `broccoli serve` to test the demo

Test 1:

- Place the cursor at the beginning of the first line, press `enter` 
- nothing happens
- Place the cursor anywhere except the beginning of the second line, press `enter` 
- the content of the line is duplicated and a new line is added

Test 2:

- Place the cursor at the beginning of the first line, press `enter`
- nothing happens
- Place the cursor anywhere except the beginning of the second line, press `enter`
- a new line is entered and no content is duplicated

Test 3: 

- Place the cursor at the beginning of the first line, press `enter` multiple times
- Place the cursor at the beginning of the second line, press `enter` multiple times
- Place the cursor anywhere except the beginning of a line, press `enter`
- Content is duplicated on both lines!

bantic:
Handle newline at start of section

  * simplify section#splitMarker
  * tests for hitting enter in middle and end of section
  * add test for hitting enter in first section
  * fix bug with PostNodeBuilder not assigning a builder to the blank marker

fixes #39

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -442,33 +442,19 @@ class Editor {
     const markerRenderNode = leftRenderNode;
     const marker = markerRenderNode.postNode;
     const section = marker.section;
-    const newMarkers = marker.split(leftOffset);
 
-    // FIXME rightMarker is not guaranteed to be there
-    let [leftMarker, rightMarker] = newMarkers;
+    let [beforeSection, afterSection] = section.splitAtMarker(marker, leftOffset);
 
-    section.markers.insertAfter(leftMarker, marker);
-    markerRenderNode.scheduleForRemoval();
+    section.renderNode.scheduleForRemoval();
 
-    const newSection = this.builder.createMarkupSection('p');
-    newSection.markers.append(rightMarker);
-
-    let nodeForMove = markerRenderNode.next;
-    while (nodeForMove) {
-      nodeForMove.scheduleForRemoval();
-      let movedMarker = nodeForMove.postNode.clone();
-      newSection.markers.append(movedMarker);
-
-      nodeForMove = nodeForMove.next;
-    }
-
-    const post = this.post;
-    post.sections.insertAfter(newSection, section);
+    this.post.sections.insertAfter(beforeSection, section);
+    this.post.sections.insertAfter(afterSection, beforeSection);
+    this.post.sections.remove(section);
 
     this.rerender();
     this.trigger('update');
 
-    this.cursor.moveToSection(newSection);
+    this.cursor.moveToSection(afterSection);
   }
 
   hasSelection() {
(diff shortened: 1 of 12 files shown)


src/js/editor/editor.js
```diff
@@ -444,17 +444,26 @@ class Editor {
...
+      afterMarker = beforeMarker;
+      beforeMarker = null;
+    }
+
+    if (beforeMarker) {
+      section.markers.insertAfter(beforeMarker, marker);
+    } else {
+      // start of section, remove the existing marker and add a blank one
+      section.markers.remove(marker);
+      section.markers.append(this.builder.createBlankMarker());
```
mixonic: Not needed, but alternatively:

```
section.markers.splice(marker, 1, [this.builder.createBlankMarker()]);
```

mixonic: Oh I'm sorry, I misread this. You're removing a marker from some arbitrary place and only appending the blank marker, all the time.

mixonic: I'm confused- shouldn't the blank marker be in a new section?

bantic: Yeah, agreed, this is fairly confusing. I'm reworking this to be a bit more declarative.

bantic:
hold off on merging for the moment. I'm reworking `Marker.split` and the way `Section` uses it, which will hopefully make this a simpler operation to code (and the code simpler to read).

bantic:
@mixonic this is ready now if you want to look it over

src/js/models/markup-section.js
```diff
@@ -53,16 +58,57 @@ export default class Section extends LinkedItem {
...
+
+  /**
+   * Remove extranous empty markers, adding one at the end if there
+   * are no longer any markers
+   *
+   * Mutates this section's markers
+   */
+  coalesceMarkers() {
+    forEach(
+      filter(this.markers, m => m.empty()),
```
mixonic: `m.empty` should probably be `m.isEmpty`

Also doing a filter only to pass it to another iterator seems wasteful- this could be done in a single pass with an `if`

mixonic: changed the loop on master, we can ponder `empty`

bantic: @mixonic removing markers while iterating them causes the iteration to skip (`item.next` is something different before/after the callback). I have a PR coming to fix iteration of the linkedlist

PR #60, opened 2015-08-11
isEmpty for linked list

mixonic:
@bantic for you creates an `isEmpty` property on the linked list instead of the `empty()` function call.


bantic:
thanks. I will update `Marker` to have an `isEmpty` property as well

Issue #57, opened 2015-08-11
Initializing with no mobiledoc dies after entering 2 characters


rlivsey:
I've initialized the editor without an initial `mobiledoc` as follows:

```javascript
let editor = new ContentKit.Editor(element, {});
```

All looks fine on the first keypress, but as soon as the second one is entered I get:

<img width="1346" alt="screenshot 2015-08-11 14 59 14" src="https://cloud.githubusercontent.com/assets/3740/9199303/a011c334-4039-11e5-8a51-e71bf65947f4.png">

The `element` here appears to be the `document` itself, so it looks like it's walking up out of the editor and not stopping until it hits the root.

It seems to be that `collectMarkups` is [being called](https://github.com/bustlelabs/content-kit-editor/blob/b4efbef396d84736b9465315e6f29f061875957c/src/js/parsers/post.js#L58-L70) with the `textNode` and `rootNode` being the same and then immediately calls `currentNode = textNode.parentNode` and so has now escaped the editor.

Should this check if `textNode === rootNode` first, or is the problem higher up?

bantic:
@rlivsey Thanks for the report — the `element` in the `Editor` constructor should be something that you choose deliberately. Can you explain what you mean when you say it appears to be the `document`?

The `element` passed to the constructor should be a dom node directly, as well — the result of a `document.getElementById` call.

rlivsey:
Here's the exact code I've got here, it's a shiny new Ember app using 1.13.5 using content-kit-editor which is npm linked into the project from a local build of master. Running on Safari & Chrome both have the same error.

```javascript
// app/components/content-editor/component.js
import Ember from 'ember';

export default Ember.Component.extend({
  didInsertElement() {
    const element = this.$(".editor")[0];
    this.editor = new ContentKit.Editor(element, {});
  },

  willDestroyElement() {
    this.editor.destroy();
  }
});
```

```handlebars
<!-- app/components/content-editor/template.hbs -->
<div class="editor"></div>
```

```handlebars
<!-- app/templates/application.hbs -->
{{content-editor}}
```

Running that and then entering two characters causes the error.

[`normalizeTagName`](https://github.com/bustlelabs/content-kit-editor/blob/c404d3b88871c6920690352392feb297521a9521/src/js/utils/dom-utils.js#L120) is called with `undefined`, because `document` doesn't have a tag name.

[...]

PR #66, opened 2015-08-12
Ensure parsed mobiledocs have a blank marker

mixonic:
This allows a valid mobiledoc with zero sections (or one section with zero markers) to be parsed into an abstract with a text prompt.

It does not address https://github.com/bustlelabs/content-kit-editor/issues/57, but a fix for that should be similar. Spiking on it now, will push here.


mixonic:
@bantic this should fix #57, handling blank mobiledoc documents and blank initial DOM nodes.

bantic:
:+1: merge at will

PR #67, opened 2015-08-12
Correct the browser's reported selection

### LINKED ISSUE #56: selecting using keyboard up arrow generates exception
The `Cursor` model currently expects the focus/anchor node in the selection to always be a text node, which it uses to lookup the associated `Marker`.

Using the `shift` key plus arrow keys can cause the browser to report selected `anchorNode`s or `focusNode`s that are not text nodes, which will cause an exception in the `Cursor#offsets` code.

On Chrome and Safari, assuming two sections, A and B:

  * Place cursor at start of section B, hold `shift`, press :arrow_up:
    * left node is a text node in the previous section A (correct)
    * right node is the paragraph tag for section B (incorrect) — should be the last text node in section A
  * Place cursor at start of section A, hold `shift`, press :arrow_down: 
    * left node is the first text node in section A (correct)
    * right node is the paragraph tag for section B (incorrect) — should be the last text node in section A

Firefox has a different issue — it also reports selected text nodes, but sometimes it reports the text node in the prev/next section instead of the correct one.

bantic:
Fixes a problem where, by using shift + arrow keys, the browser would report
a selection of an entire selection rather than two text nodes.

fixes #56

### DIFF
--- a/src/js/models/cursor.js
+++ b/src/js/models/cursor.js
@@ -8,8 +8,32 @@ import {
 } from '../utils/selection-utils';
 
 import {
-  detectParentNode
+  detectParentNode,
+  isTextNode
 } from '../utils/dom-utils';
+
+
+function comparePosition(selection) {
+  let { anchorNode, focusNode, anchorOffset, focusOffset } = selection;
+  let leftNode, rightNode, leftOffset, rightOffset;
+
+  const position = anchorNode.compareDocumentPosition(focusNode);
+
+  if (position & Node.DOCUMENT_POSITION_FOLLOWING) {
+    leftNode = anchorNode; rightNode = focusNode;
+    leftOffset = anchorOffset; rightOffset = focusOffset;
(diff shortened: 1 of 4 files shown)


mixonic:
gnarly.

src/js/models/cursor.js
```diff
@@ -8,9 +8,33 @@ import {
...
+
+function comparePosition(selection) {
+  let { anchorNode, focusNode, anchorOffset, focusOffset } = selection;
+  let leftNode, rightNode, leftOffset, rightOffset;
+
+  const position = anchorNode.compareDocumentPosition(focusNode);
+
+  if (position & Node.DOCUMENT_POSITION_FOLLOWING) {
+    leftNode = anchorNode; rightNode = focusNode;
+    leftOffset = anchorOffset; rightOffset = focusOffset;
```
mixonic: Whoa, many statements on one line. Unexpected.

PR #69, opened 2015-08-12
Render a last marker with trailing space using NO_BREAK_SPACE

### LINKED ISSUE #68: trailing space in section is eaten by browser when deleting
In Chrome and Safari, when the cursor is at a character with a space behind it, after hitting delete once the cursor skips backward across the space:

  * text is "abc d", cursor is after the d.
  * Hit delete and the cursor is now after the "c" instead of after the space. The DOM still has the space but you can't use the arrow keys to move the cursor past it.

Firefox leaves the cursor after the space, as expected.

bantic:
If the text node ends with a NO_BREAK_SPACE character ([U+00A0](http://www.fileformat.info/info/unicode/char/00a0/index.htm)),
using `selection.addRange` will display the character at the proper location,
and typing new text will correctly put it after the trailing space.

Fixes #68

cc @mixonic

### DIFF
--- a/src/js/renderers/editor-dom.js
+++ b/src/js/renderers/editor-dom.js
@@ -9,6 +9,8 @@ import { CARD_TYPE } from "../models/card";
 import { clearChildNodes } from '../utils/dom-utils';
 
 export const UNPRINTABLE_CHARACTER = "\u200C";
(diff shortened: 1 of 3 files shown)


mixonic:
:100: :100: Gracias.

PR #72, opened 2015-08-12
add `LinkedList#removeBy`

bantic:
If iterating a list and removing items, the `forEach` will skip the
item after the current one.

See also https://github.com/bustlelabs/content-kit-editor/pull/53#discussion_r36903458

cc @mixonic


mixonic:
We discussed writing a mutating `removeBy` as an alternative to this. I'm uncomfortable that the change to `forEach` would diverge it from the semantics on the corresponding array method.

PR #73, opened 2015-08-12
Refactor some method into public postEditor methods

mixonic:
This begins the exposure of a public API programmatically interacting with the editor. For example:

```js
let editor = new ContentKit.Editor(someElement);

$('button').append(document.body).on('click', () => {

  let markerRange = editor.cursor.offsets;
  let markup = editor.builder.createMarkup('strong');
  editor.run((postEditor) => {
    postEditor.removeMarkupFromMarkers(markerRange, markup);
  });

});
```

While the editor and the cursor may create post abstract objects or read data about the current state of the editor, any modifications must be make through the `postEditor` interface. The runloop of the `postEditor` ensures that no matter what operations are performed only a single render and update are performed.

TODO:
 
* [x] On the `postEditor`, `this.rerender()` should be `this.scheduleRerender()`. Likewise for `didUpdate`
* [x] Docs
* [x] The internals of `handleNewline` must be migrated to `postEditor`
* [ ] ~~Operations during `reparse` must move to the `postEditor`~~ actually, reparse should be going away now that we control delete/enter. Should discuss with @bantic.
* [ ] ~~`loadModel` should become part of the editor~~ dropped loadModel, we don't really have an immediate need.
* [ ] ~~`rerender` itself should be moved to the `postEditor`, and the renderer should be passed to the `postEditor` instance.~~ seems coupled to loadModel concerns. Will revisit.

### DIFF
--- a/src/js/commands/italic.js
+++ b/src/js/commands/italic.js
@@ -15,6 +15,9 @@ export default class ItalicCommand extends TextFormatCommand {
   }
   exec() {
     let markerRange = this.editor.cursor.offsets;
+    if (!markerRange.leftRenderNode || !markerRange.rightRenderNode) {
+      return;
+    }
     let markers = this.editor.run((postEditor) => {
       return postEditor.applyMarkupToMarkers(markerRange, this.markup);
     });
--- a/src/js/commands/italic.js
+++ b/src/js/commands/italic.js
@@ -14,10 +14,18 @@ export default class ItalicCommand extends TextFormatCommand {
     this.markup = builder.createMarkup('em');
   }
   exec() {
-    this.editor.applyMarkupToSelection(this.markup);
+    let markerRange = this.editor.cursor.offsets;
+    let markers = this.editor.run((postEditor) => {
+      return postEditor.applyMarkupToMarkers(markerRange, this.markup);
+    });
+    this.editor.selectMarkers(markers);
   }
   unexec() {
-    this.editor.removeMarkupFromSelection(this.markup);
+    let markerRange = this.editor.cursor.offsets;
+    let markers = this.editor.run((postEditor) => {
+      return postEditor.removeMarkupFromMarkers(markerRange, this.markup);
+    });
+    this.editor.selectMarkers(markers);
   }
   isActive() {
     return any(this.editor.activeMarkers, m => m.hasMarkup(this.markup));
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -206,6 +206,20 @@ class PostEditor {
     return markers;
   }
 
+  insertSectionBefore(section, beforeSection) {
+    this.editor.post.sections.insertBefore(section, beforeSection);
+    this.editor.post.renderNode.markDirty();
+
+    this.rerender();
+    this.didUpdate();
+  }
+
+  removeSection(section) {
+    section.renderNode.scheduleForRemoval();
+
+    this.rerender();
+    this.didUpdate();
+  }
 
   /**
    * A method for adding work the deferred queue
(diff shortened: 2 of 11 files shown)


mixonic:
Rebase and merge with @bantic's latest, also handling of newlines is ported. Moving through the other items.

tests/unit/models/markup-section-test.js
```diff
@@ -149,3 +149,21 @@ test('#coalesceMarkers appends a single blank marker if all the markers were bla
   assert.equal(s.markers.length, 1, 'has 1 marker after coalescing');
   assert.ok(s.markers.head.isEmpty, 'remaining marker is empty');
 });
+
+test('.isBlank returns true if the text length is zero for two markers', (assert) => {
```
bantic: Elsewhere in the tests we use `#` for instance method abbreviation and `.` (dot) for class method abbreviations. To stay in sync with those this should say `"#isBlank returns ..."`

tests/unit/models/markup-section-test.js
```diff
@@ -149,3 +149,21 @@ test('#coalesceMarkers appends a single blank marker if all the markers were bla
...
+  const m1 = builder.createBlankMarker();
+  const m2 = builder.createBlankMarker();
+  const s = builder.createMarkupSection('p', [m1,m2]);
+  assert.ok(s.isBlank, 'section with two blank makers is blank');
```
bantic: "maker" -> "markers" (also in the test below this)

src/js/models/cursor.js
```diff
@@ -123,16 +129,16 @@ export default class Cursor {
...
-    let r = document.createRange();
-    r.selectNode(markerElement);
-    r.collapse(true);
-    const selection = this.selection;
-    if (selection.rangeCount > 0) {
-      selection.removeAllRanges();
-    }
-    selection.addRange(r);
+  // moves cursor to marker
+  moveToMarker(headMarker, headOffset=0, tailMarker=headMarker, tailOffset=headOffset) {
```
bantic: nice

src/js/commands/italic.js
```diff
@@ -14,10 +14,21 @@ export default class ItalicCommand extends TextFormatCommand {
     this.markup = builder.createMarkup('em');
   }
   exec() {
-    this.editor.applyMarkupToSelection(this.markup);
+    let markerRange = this.editor.cursor.offsets;
+    if (!markerRange.leftRenderNode || !markerRange.rightRenderNode) {
+      return;
+    }
```
bantic: This feels like inside baseball. The italic command should never be `exec`-able if there isn't a selection — if the right/left render nodes are not present then something else is wrong, or the command is being `exec`'d when there's no selection to apply markup to.

src/js/editor/post.js
```diff
@@ -0,0 +1,474 @@
...
+   *     });
+   *
+   * @method insertSectionBefore
+   * @param {Object} section The new section
+   * @param {Object} beforeSection The section "before" is relative to
+   * @public
+   */
+  insertSectionBefore(section, beforeSection) {
+    this.editor.post.sections.insertBefore(section, beforeSection);
+    this.editor.post.renderNode.markDirty();
```
bantic: does the entire post need to be marked dirty? Can we only dirty the section and beforeSection?

bantic:
:+1: this is great

src/js/commands/italic.js
mixonic: Ah. I was able to create a failure here in chrome with some wacky selection drags. We should look at it together, I'll make a ticket.

src/js/editor/post.js
mixonic: The post must be marked dirty so the new section be discovered. But the render will be crazy cheap, b/c on the two dirty sections will actually be rendered. All the others just sit still.

PR #74, opened 2015-08-12
Make getUserMedia work across browsers

toddself:
No need to make firefox users unable to take selfies! Am looking into
why the mobiledoc output box is empty on Firefox as well.


ZeeJab:
@toddself hahah yessss!!!

mixonic:
Amazing :-) Thanks @toddself

PR #82, opened 2015-08-17
Add disableEditing, enableEditing

mixonic:
Per the [0.4.0 epic](https://github.com/bustlelabs/content-kit-editor/issues/34), we would like an API for disabling the ability for a user to enter and edit existing text. This is useful in conjunction with the programmatic API- It allows us to build editing interfaces that output mobiledoc and cards, but without needing to allow freeform content.

The APIs are:

* `editor.disableEditing()` - stop the user from being able to edit the
  current post with their cursor. Programmatic edits are still allowed.
* `editor.enableEditing()` - allow the user to make direct edits directly
  to a post's text.

### DIFF
--- a/tests/acceptance/editor-sections-test.js
+++ b/tests/acceptance/editor-sections-test.js
@@ -167,7 +167,22 @@ test('hitting enter at end of a section creates new empty section', (assert) => 
                     offset: 0});
 });
 
-test('deleting across 0 sections merges them', (assert) => {
+// Phantom does not recognize toggling contenteditable off
+Helpers.skipInPhantom('deleting across 2 sections does nothing if editing is disabled', (assert) => {
+  editor = new Editor(editorElement, {mobiledoc: mobileDocWith2Sections});
+  editor.disableEditing();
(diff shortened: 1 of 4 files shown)


mixonic:
@bantic could use a review here!

tests/acceptance/editor-sections-test.js
```diff
@@ -167,6 +167,20 @@ test('hitting enter at end of a section creates new empty section', (assert) =>
                     offset: 0});
 });
 
+Helpers.skipInPhantom('deleting across 0 sections does nothing if editing is disabled', (assert) => {
```
bantic: this should be safe to run in phantom, and should say "delete across 2 sections"

bantic:
@mixonic :+1: I had a small comment on one of the tests. It would be nice if there were a single bottleneck that event listeners ran through — that way there'd only be a single point where you would ignore events — like a `handle(Xevent)`, but that's a separate concern. This looks good

mixonic:
Phantom does not respect toggling `contenteditable` off :-/ Left the skip in with a comment. Confirmed to pass locally.

PR #83, opened 2015-08-17
Add forward and backward deletion to PostEditor

### LINKED ISSUE #36: Forward delete behaves the same as backspace
Pressing the `delete` key on a full keyboard or `fn + backspace` on a mac keyboard results in the character before being deleted instead of the character afterward. 

This behaved as I would expect, deleting forwards in 0.2.1, and changed in 0.2.2

How I reproduced:

- on master (58ffbc8) using `broccoli serve` to test the demo
- place the cursor between `h` and `2` on the first line and press `delete`
- the character `h` is removed, instead of `2`

bantic:
Changes `postEditor#deleteCharAt` to be `postEditor#deleteFrom(position, direction)`.

Adds unit tests for the `postEditor` directly that assert that it changes
the AT as expected.

Some cleanup in the editor as well to remove some duplicated code.

fixes #36

cc @mixonic

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -238,10 +238,12 @@ class Editor {
     element.setAttribute('contentEditable', true);
 
     if (this.mobiledoc) {
-      this.parseModelFromMobiledoc(this.mobiledoc);
+      this.post = new MobiledocParser(this.builder).parse(this.mobiledoc);
     } else {
-      this.parseModelFromDOM(this.element);
+      this.post = new DOMParser(this.builder).parse(this.element);
     }
+
+    this._renderTree = this.prepareRenderTree(this.post);
 
     clearChildNodes(element);
     this.rerender();
@@ -276,21 +278,11 @@ class Editor {
     this._views.push(view);
   }
 
-  parseModelFromDOM(element) {
-    let parser = new DOMParser(this.builder);
-    this.post = parser.parse(element);
-    this._renderTree = new RenderTree();
-    let node = this._renderTree.buildRenderNode(this.post);
-    this._renderTree.node = node;
-    this.trigger('update');
-  }
-
-  parseModelFromMobiledoc(mobiledoc) {
-    this.post = new MobiledocParser(this.builder).parse(mobiledoc);
-    this._renderTree = new RenderTree();
-    let node = this._renderTree.buildRenderNode(this.post);
-    this._renderTree.node = node;
-    this.trigger('update');
+  prepareRenderTree(post) {
+    let renderTree = new RenderTree();
+    let node = renderTree.buildRenderNode(post);
+    renderTree.node = node;
+    return renderTree;
   }
 
   rerender() {
@@ -308,20 +300,23 @@ class Editor {
   handleDeletion(event) {
     event.preventDefault();
 
-    let offsets = this.cursor.offsets;
-    let currentMarker, currentOffset;
-    this.run((postEditor) => {
-      let results;
+
+    const results = this.run(postEditor => {
+      const offsets = this.cursor.offsets;
+
       if (this.cursor.hasSelection()) {
-        results = postEditor.deleteRange(offsets);
+        return postEditor.deleteRange(offsets);
       } else {
-        // FIXME: perhaps this should accept this.cursor.offsets?
-        results = postEditor.deleteCharAt(offsets.headMarker, offsets.headOffset-1);
+        const {headMarker, headOffset} = offsets;
+        const key = Key.fromEvent(event);
+
+        const deletePosition = {marker: headMarker, offset: headOffset},
+              direction = key.direction;
+        return postEditor.deleteFrom(deletePosition, direction);
       }
-      currentMarker = results.currentMarker;
-      currentOffset = results.currentOffset;
     });
-    this.cursor.moveToMarker(currentMarker, currentOffset);
+
+    this.cursor.moveToMarker(results.currentMarker, results.currentOffset);
   }
 
   handleNewline(event) {
--- a/tests/helpers/mobiledoc.js
+++ b/tests/helpers/mobiledoc.js
@@ -1,0 +1,28 @@
+import PostNodeBuilder from 'content-kit-editor/models/post-node-builder';
+import MobiledocRenderer from 'content-kit-editor/renderers/mobiledoc';
+
+/*
+ * usage:
+ *  makeMD(({post, section, marker, markup}) =>
+ *    post([
+ *      section('P', [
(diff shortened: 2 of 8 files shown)


src/js/editor/editor.js
```diff
@@ -238,11 +238,13 @@ class Editor {
...
 
     if (this.mobiledoc) {
-      this.parseModelFromMobiledoc(this.mobiledoc);
+      this.post = new MobiledocParser(this.builder).parse(this.mobiledoc);
     } else {
-      this.parseModelFromDOM(this.element);
+      this.post = new DOMParser(this.builder).parse(this.element);
     }
 
+    this._renderTree = this.prepareRenderTree(this.post);
```
mixonic: :+1:

src/js/editor/editor.js
```diff
@@ -307,20 +299,27 @@ class Editor {
...
       if (this.cursor.hasSelection()) {
         results = postEditor.deleteRange(offsets);
       } else {
-        // FIXME: perhaps this should accept this.cursor.offsets?
-        results = postEditor.deleteCharAt(offsets.headMarker, offsets.headOffset-1);
+        const deletePosition = {
+          marker: offsets.headMarker,
+          offset: offsets.headOffset
+        };
+        results = postEditor.deleteFrom(deletePosition, direction);
```
mixonic: Hm, if we add a `deleteFrom` can it just be added on top of a `deleteCharAt` API? I imagine a generic character deletion API would be useful.

Or should the `deleteRange` just be used. Could be build both a `deleteCharAt` and `deleteFrom` on top of `deleteRange`? I found having all the logic for a given case flowing through one function was really useful for the linked list, ala `insertBefore` managing append, prepend, insertAfter etc.

tests/helpers/mobiledoc.js
```diff
@@ -0,0 +1,26 @@
...
+       section('P', [
+         marker('some text', [markup('B')])
+       ])
+     })
+ */
+function makeMD(treeFn) {
+  let builder = new PostNodeBuilder();
+
+  const post    = (...args) => builder.createPost(...args);
+  const section = (...args) => builder.createMarkupSection(...args);
```
mixonic: should be `markupSection`. There are multiple types of sections, and calling this one `section` leads to the others being disregarded. :neckbeard:

tests/helpers/mobiledoc.js
```diff
@@ -0,0 +1,26 @@
...
+
+/*
+ * makeMD({post, section, marker, markup} =>
+     post([
+       section('P', [
+         marker('some text', [markup('B')])
+       ])
+     })
+ */
+function makeMD(treeFn) {
```
mixonic: Can this be `makeMobiledoc`?

Additionally, those docs do not look like valid syntax.

src/js/editor/editor.js
bantic: Yeah, I think directional deletion makes something like `deleteCharAt` feel awkward. If we are deleting, the direction matters. And we want to delete __from__ a cursor position rather than __at__ a character offset (that's why I changed the method name from `deleteCharAt` to `deleteFrom`). For instance to backward-delete "__at__" the beginning of a section (joining with the previous section) is semantically different than deleting the character "at" the 0th position of the section (in that case, at least to my ears, no joining would be implied, just removal of the character).

I could see an argument for a `deleteCharAt` method that had no implication of joining anything and only removes a character in a marker at a given character offset. I'd consider `removeCharAt(marker, offset)` to be a better name, though.

I still have a little trouble following the many conditionals of LL's `insertBefore` — I would be hesitant to combine the forward- and backward- deletion cases into a single method for fear that the logic would be tortuous. They have a lot of symmetric similarity, but only a little duplicated code (when the cursor offset is in the middle of a marker — that would be a good candidate to refactor into a `removeCharAt` method), and there's a real difference between them in that the `nextCursor` marker and offset do not need to change (except one edge case) for forward-deletion but they do in nearly every case for backward-deletion.

tests/helpers/mobiledoc.js
bantic: ha, I'll change it. I always consider card sections to be special types of sections. I'm going to have to make a vim macro to cut down on the typing. :)

tests/helpers/mobiledoc.js
bantic: Maybe change it to `build` or just `make`? It is used by `Helpers.mobiledoc` as `Helpers.mobiledoc.makeMD`, already slightly redundant. Changing to `makeMobiledoc` would make the code read `Helpers.mobiledoc.makeMobiledoc`.

The mobiledoc should be valid — it first builds a post and then converts it to mobiledoc.

src/js/editor/editor.js
mixonic: These all seem like different views of deleting a range to me. Backward 
should ideally be as simple as subtracting a position from the offsets.

As for the unified logic, I don’t think that should be take to mean 
“use one big function”. Instead I think it means to build for one 
master format of inputs. The semantic uses are then wrappers that adapt 
to that master format, and the internals of that format we then know to 
be pretty safe for all cases. The alternative is that you’re 
implementing the same deletion logic several times, and likely not 
covering all opportunities for failure.

Ideally the cursor doesn’t play into the data model changes. I’m 
aware it does right now, in that a currentMarker and offset are 
returned.

tests/helpers/mobiledoc.js
mixonic: `build` seems fine to me. I'm just railing against abbreviations.

```js
makeMD({post, section, marker, markup} =>
  post([
    section('P', [
      marker('some text', [markup('B')])
    ])
  })
```
is not valid JS. I ported it to something that parses:

```js
makeMD(({post, section, marker, markup}) => {
  return post([
    section('P', [
      marker('some text', [markup('B')])
    ])
  ]);
});
```

bantic: I see — fixed. And changed to `build`

src/js/editor/editor.js
bantic: I think when we have a lot of potentially different input formats floating around we're likely missing an abstraction. I've been wanting a `Position` or `CursorPosition` container that makes passing around positional information easier. The `deleteFrom({marker, offset})` could become `deleteFrom(positionInstance)` in the future.

Having a position abstraction would open up `deleteRange` so that we could do things like `deleteRange(from=cursorPosition, to=cursorPosition.moveBackOne())` and `deleteRange(from=cursorPosition, to=cursorPosition.moveForwardOne())`. I would like to tackle that in a future PR, though. There is good test coverage for `deleteFrom` now that we can lean on until then.

PR #89, opened 2015-08-25
Split render from editor instantiation

mixonic:
Drop the element argument from editor instantiation.

Should this perhaps be `editor.append`?

Fixes https://github.com/bustlelabs/content-kit-editor/issues/88

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -30,7 +30,8 @@ import MobiledocRenderer from '../renderers/mobiledoc';
 import { mergeWithOptions } from 'content-kit-utils';
 import {
   clearChildNodes,
-  addClassName
+  addClassName,
+  parseHTML
 } from '../utils/dom-utils';
 import {
   forEach,
@@ -69,7 +70,8 @@ const defaults = {
   unknownCardHandler: () => {
     throw new Error('Unknown card encountered');
   },
-  mobiledoc: null
+  mobiledoc: null,
+  html: null
 };
 
 function bindContentEditableTypingListeners(editor) {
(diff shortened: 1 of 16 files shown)


src/js/editor/editor.js
```diff
@@ -71,7 +71,8 @@ const defaults = {
   unknownCardHandler: () => {
     throw new Error('Unknown card encountered');
   },
-  mobiledoc: null
+  mobiledoc: null,
+  pasteInput: null
```
bantic: `html` may be a more obvious name for this. `pasteInput` implies a that is the result of a copy-paste operation (which needn't be true) and is vague about what the `input` is (text, html, dom node, etc).

bantic:
Was adding an option like `renderImmediately: false` considered? Using an extra call to `render` seems like it may come as a mild surprise to an implementer; the 80% use case seems like it would be to instantiate an editor and immediately render it.
Otherwise :+1:

src/js/editor/editor.js
mixonic: `dom` felt pretty lame. Bleh, I guess it can be `html` and I can make it recognize when DOM is passed instead.

mixonic:
I'd prefer one flow over making immediate render the default case. IMO Having a default render behavior encourages us to be lazy about what things look like before rendering has occurred. Several APIs should fail if you have not yet rendered.

Having the editor pre-render allows the user to call configuration methods like `disableInput` without us also needing to translate those into options.

mixonic:
@bantic let us continue to discuss. I'm going to merge and cut a release to move forward though.

Issue #86, opened 2015-08-18
implement lists


bantic:
A list section is different from other markup sections because it has sub-elements that are not markers.

A normal markup section has (a linked-list of) markers:
`markupSection -> markers`

But a list section has list items which have markers:
`listSection -> listItems -> markers`

To accommodate lists, the Abstract Tree (AT) will need to be able to traverse the leaf nodes (markers) of the tree — for semantic actions like splitting/joining sections and markers — without assuming that the max depth of children is always going to be 1.

  * [x] editor-dom renderer can render list sections and list items
  * [x] mobiledoc parser can parse list sections and list items
  * [ ] mobiledoc dom renderer can render list sections and items
  * [ ] mobiledoc html renderer can render list sections and items
  * [x] Post parser must be able to reparse list sections
  * [x] Actions on the post editor like `splitMarkers` must be able to handle markers that have a depth > 1

Creating a list will happen the same way that other section types are created: Click a toolbar button. Eventually (#87) also via text-expansion.
Exiting a list happens by hitting enter when the cursor is in the last item in a list and it is empty (this is how google docs works, too).

For now, only singly-nested lists (this is how medium does it).

bantic:
WIP on the [lists](https://github.com/bustlelabs/content-kit-editor/compare/lists) branch

PR #90, opened 2015-08-25
Remove UNPRINTABLE_CHARACTER, use <br> instead

bantic:
* use position via cursor offset in section, rather than marker
  * add section#markersFrom

A lot of small refactorings inside. Aside from using a `<br>` instead of the unprintable zero-width non-joiner space, we also return offsets from the start of the section in `cursor.offsets`

cc @mixonic

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -24,9 +24,6 @@ import MobiledocParser from "../parsers/mobiledoc";
 import PostParser from '../parsers/post';
 import DOMParser from '../parsers/dom';
 import Renderer  from 'content-kit-editor/renderers/editor-dom';
-import {
-  UNPRINTABLE_CHARACTER
-} from 'content-kit-editor/renderers/editor-dom';
 import RenderTree from 'content-kit-editor/models/render-tree';
 import MobiledocRenderer from '../renderers/mobiledoc';
 
@@ -36,7 +33,8 @@ import {
   addClassName
 } from '../utils/dom-utils';
 import {
-  forEach
+  forEach,
+  detect
 } from '../utils/array-utils';
 import { getData, setData } from '../utils/element-utils';
 import mixin from '../utils/mixin';
@@ -157,10 +155,11 @@ function bindKeyListeners(editor) {
       editor.handleNewline(event);
     } else if (key.isPrintable()) {
       if (editor.cursor.hasSelection()) {
-        let result = editor.run((postEditor) => {
-          return postEditor.deleteRange(editor.cursor.offsets);
+        let offsets = editor.cursor.offsets;
+        editor.run((postEditor) => {
+          postEditor.deleteRange(editor.cursor.offsets);
         });
-        editor.cursor.moveToMarker(result.currentMarker, result.currentOffset);
+        editor.cursor.moveToSection(offsets.headSection, offsets.headSectionOffset);
       }
     }
   });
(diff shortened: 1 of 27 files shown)


bantic:
This work resulted from some exploration related to implementing lists (#86). Changing to use a br seems to be the most cross-browser-safe way to programmatically position a cursor in an blank space. using the zero-width non-joiner space created difficulties around cursor movement (the cursor stays in the same visual place when moving "past" it with left/right arrow, for instance).

src/js/editor/editor.js
```diff
@@ -157,10 +155,11 @@ function bindKeyListeners(editor) {
...
     } else if (key.isPrintable()) {
       if (editor.cursor.hasSelection()) {
-        let result = editor.run((postEditor) => {
-          return postEditor.deleteRange(editor.cursor.offsets);
+        let offsets = editor.cursor.offsets;
+        editor.run((postEditor) => {
+          postEditor.deleteRange(editor.cursor.offsets);
         });
-        editor.cursor.moveToMarker(result.currentMarker, result.currentOffset);
+        editor.cursor.moveToSection(offsets.headSection, offsets.headSectionOffset);
```
mixonic: We should just add a `restoreOffset` that takes the offset. It would need to error when the expected sections were no longer present, but would be really useful besides then.

bantic: you mean `editor.cursor.restoreOffset(offset)` ? How would it know which section?

src/js/models/markup-section.js
```diff
@@ -115,10 +116,73 @@ export default class Section extends LinkedItem {
...
+        if (!afterMarker) {
+          afterMarker = m;
+        }
+        didAddContent = true;
+      }
+    });
+
+    if (wasBlank && didAddContent) {
+      // FIXME: Join should maybe not be on the markup-section
+      beforeMarker.renderNode.scheduleForRemoval();
```
mixonic: This is a notable flaw we should resolve in a followup.

src/js/models/markup-section.js
```diff
@@ -115,10 +116,73 @@ export default class Section extends LinkedItem {
...
+
+  markersFor(headOffset, tailOffset) {
+    let markers = [];
+    let adjustedHead = 0, adjustedTail = 0;
+    this.markers.forEach(m => {
+      adjustedTail += m.length;
+
+      if (adjustedTail > headOffset && adjustedHead < tailOffset) {
+        let head = Math.max(headOffset - adjustedHead, 0);
+        let tail = adjustedTail < tailOffset ? m.length : m.length - (adjustedTail - tailOffset);
```
mixonic: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes:

src/js/models/post.js
```diff
@@ -5,60 +5,73 @@ export default class Post {
...
-    changedSections.forEach(section => {
-      if (section.markers.isEmpty) {
-        section.markers.append(this.builder.createBlankMarker());
-      }
-    });
+    return foundMarkers;
+  }
 
-    let currentMarker, currentOffset;
+  cutMarkers(markers) {
```
mixonic: Looking forward to refactoring this.

src/js/editor/editor.js
mixonic: The offset contains the section.

```
let offsets = this.cursor.offsets;
editor.cursor.restoreOffsets(offsets); // Would throw if the sections are no longer valid.
```

src/js/models/markup-section.js
bantic: :+1: I actually realized it's even simpler:
`let tail = m.length - Math.max(adjustedTail - tailOffset, 0);`
fixed on master

PR #91, opened 2015-08-25
Ensure that editor gets destroyed between tests

bantic:
cc @mixonic

These tests scoped their own `editor` variable and so the editor teardown
next happened. This leaks listeners from attached views (notably EmbedIntent).
If you click on the test page after the tests are done running you can see
EmbedIntent's click handler continuing to fire (and generating errors).


mixonic:
Thanks, good catch!

PR #93, opened 2015-08-26
Fix clicking when there is no active cursor.

mixonic:
Also refactors the `offsets` return value into a `Range` object consisting of two `Position` objects. These encapsulate the difficulties of creating a meaningful position data-set for a given DOM node.

### DIFF
--- a/src/js/utils/cursor.js
+++ b/src/js/utils/cursor.js
@@ -1,17 +1,15 @@
-import {
-  detect
-} from '../utils/array-utils';
-
+import { detect } from '../utils/array-utils';
 import {
   isSelectionInElement,
   clearSelection
 } from '../utils/selection-utils';
-
 import {
   detectParentNode,
   isTextNode,
   walkDOM
 } from '../utils/dom-utils';
+import Position from "./cursor/position";
+import Range from "./cursor/range";
 
 function findOffsetInParent(parentElement, targetElement, targetOffset) {
   let offset = 0;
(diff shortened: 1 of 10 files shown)


mixonic:
Also moved `cursor` into `utils`. I'd like to get to a point where the `models` directory is all abstract post stuff.

src/js/utils/cursor.js
```diff
@@ -98,87 +98,29 @@ export default class Cursor {
 
   get offsets() {
     const { sections } = this.post;
-    const selection = this.selection;
-    const { isCollapsed, rangeCount } = selection;
-    const range = rangeCount > 0 && selection.getRangeAt(0);
+    const { selection } = this;
 
-    if (!range) {
+    if (selection.rangeCount === 0 || !selection.getRangeAt(0)) {
```
bantic: I think `selection.rangeCount === 0` is sufficient. I'm not sure what circumstance would result in the rangeCount being non-zero and `getRangeAt(0)` returning `undefined`

src/js/utils/cursor/position.js
```diff
@@ -0,0 +1,60 @@
...
+        offsetInSection
+      );
+      marker = markerPosition.marker;
+      offsetInMarker = markerPosition.offset;
+    }
+
+    this.section = section;
+    this.offsetInSection = offsetInSection;
+    this.marker = marker;
+    this.offsetInMarker = offsetInMarker;
```
bantic: seems like we could use getters to make these calculations lazy

tests/acceptance/basic-editor-test.js
```diff
@@ -62,3 +57,27 @@ test('#disableEditing and #enableEditing toggle contenteditable', (assert) => {
...
 });
+
+test('clicking outside the editor does not raise an error', (assert) => {
+  editor = new Editor({autofocus: false});
+  editor.render(editorElement);
+
+  let secondEditorElement = document.createElement('div');
+  document.body.appendChild(secondEditorElement);
+
+  let secondEditor = new Editor(); // This editor will be focused
```
bantic: could make the comment redundant by passing `{autofocus:true}`

bantic:
:+1:

PR #94, opened 2015-08-26
Implement Lists

### LINKED ISSUE #86: implement lists

bantic:
*ready to merge*

fixes #86 

This adds a:
  * `ListSection` — inherits from `MarkupSection` and lives alongside `MarkupSection`s and cards in the post's `sections` list
  * `ListItem` — inherits from `MarkupSection` and lives in the nested `items`/`sections` list on a `ListSection`, has a list of `markers`
  * Some edge case code to `postEditor#splitSection` to deal with hitting enter on an empty final list item
  * Some largely duplicated code in `ListItem#splitAtMarker` that could be cleaned up
  * Simplifies `editor#reparse` by recognizing that there are fewer unexpected situations that could occur
  * Add a new type of section ("3") to mobiledoc format for lists
  * Update the Mobiledoc renderer to render ListSections and ListItems
  * Update the EditorDom renderer to render ListSections and ListItems
  * Update MOBILEDOC_VERSION to "0.2.0"

Stil to do:
  * [ ] update the mobiledoc html/dom renderers to recognize the change to the mobiledoc format ([dom](https://github.com/bustlelabs/mobiledoc-dom-renderer/issues/4) and [html](https://github.com/bustlelabs/mobiledoc-html-renderer/issues/4))
  * [x] clean up duplicated code
  * [x] there are still some slightly buggy behaviors around adding multiple list items, apply markup across list items, etc. Corral these

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -31,7 +31,7 @@
   "undef"         : true,     // Require all non-global variables be declared before they are used.
   "unused"        : true,     // Warn when variables are created but not used.
   "trailing"      : true,     // Prohibit trailing whitespaces.
-  "es3"           : false,     // Prohibit trailing commas for old IE 
+  "es3"           : true,     // Prohibit trailing commas for old IE 
   "esnext"        : true,     // Allow ES.next specific features such as `const` and `let`.
   
   // == Relaxing Options ================================================
--- a/.jshintrc
+++ b/.jshintrc
@@ -31,7 +31,7 @@
   "undef"         : true,     // Require all non-global variables be declared before they are used.
   "unused"        : true,     // Warn when variables are created but not used.
   "trailing"      : true,     // Prohibit trailing whitespaces.
-  "es3"           : true,     // Prohibit trailing commas for old IE 
+  "es3"           : false,     // Prohibit trailing commas for old IE 
   "esnext"        : true,     // Allow ES.next specific features such as `const` and `let`.
   
   // == Relaxing Options ================================================
--- a/src/js/commands/image.js
+++ b/src/js/commands/image.js
@@ -13,12 +13,13 @@ export default class ImageCommand extends Command {
     let beforeSection = headMarker.section;
     let afterSection = beforeSection.next;
     let section = this.editor.builder.createCardSection('image');
+    const collection = beforeSection.parent.sections;
 
     this.editor.run((postEditor) => {
       if (beforeSection.isBlank) {
         postEditor.removeSection(beforeSection);
       }
-      postEditor.insertSectionBefore(section, afterSection);
+      postEditor.insertSectionBefore(collection, section, afterSection);
     });
   }
 }
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -35,7 +35,7 @@ import {
 } from '../utils/dom-utils';
 import {
   forEach,
-  detect
+  filter
 } from '../utils/array-utils';
 import { getData, setData } from '../utils/element-utils';
 import mixin from '../utils/mixin';
@@ -446,78 +446,27 @@ class Editor {
   }
 
   reparse() {
-    // find added sections
-    let sectionsInDOM = [];
-    let newSections = [];
-    let previousSection;
+    let { headSection, headSectionOffset } = this.cursor.offsets;
+    if (headSectionOffset === 0) {
+      // FIXME if the offset is 0, the user is typing the first character
+      // in an empty section, so we need to move the cursor 1 letter forward
+      headSectionOffset = 1;
+    }
 
-    forEach(this.element.childNodes, (node) => {
-      // FIXME: this is kind of slow
-      let sectionRenderNode = detect(this._renderTree.node.childNodes, (renderNode) => {
-        return renderNode.element === node;
-      });
-      if (!sectionRenderNode) {
-        let section = this._parser.parseSection(node);
-        newSections.push(section);
-
-        // create a clean "already-rendered" node to represent the fact that
-        // this (new) section is already in DOM
-        sectionRenderNode = this._renderTree.buildRenderNode(section);
-        sectionRenderNode.element = node;
-        sectionRenderNode.markClean();
-
-        let previousSectionRenderNode = previousSection && previousSection.renderNode;
-        this.post.sections.insertAfter(section, previousSection);
-        this._renderTree.node.childNodes.insertAfter(sectionRenderNode, previousSectionRenderNode);
-      }
-
-      // may cause duplicates to be included
-      let section = sectionRenderNode.postNode;
-      sectionsInDOM.push(section);
-      previousSection = section;
-    });
-
-    // remove deleted nodes
-    const deletedSections = [];
-    forEach(this.post.sections, (section) => {
-      if (!section.renderNode) {
-        throw new Error('All sections are expected to have a renderNode');
-      }
-
-      if (sectionsInDOM.indexOf(section) === -1) {
-        deletedSections.push(section);
-      }
-    });
-    forEach(deletedSections, (s) => s.renderNode.scheduleForRemoval());
-
-    // reparse the new section(s) with the cursor
-    // to ensure that we catch any changed html that the browser might have
-    // added
-    const sectionsWithCursor = this.cursor.activeSections;
-    forEach(sectionsWithCursor, (section) => {
-      if (newSections.indexOf(section) === -1) {
-        this.reparseSection(section);
-      }
-    });
-
-    let {
-      headSection,
-      headSectionOffset
-    } = this.cursor.sectionOffsets;
-
-    // The cursor will lose its textNode if we have reparsed (and thus will rerender, below)
-    // its section. Ensure the cursor is placed where it should be after render.
-    //
-    // New sections are presumed clean, and thus do not get rerendered and lose
-    // their cursor position.
-    let resetCursor = sectionsWithCursor.indexOf(headSection) !== -1;
+    this._reparseCurrentSection();
+    this._removeDetachedSections();
 
     this.rerender();
     this.trigger('update');
 
-    if (resetCursor) {
-      this.cursor.moveToSection(headSection, headSectionOffset);
-    }
+    this.cursor.moveToSection(headSection, headSectionOffset);
+  }
+
+  _removeDetachedSections() {
+    forEach(
+      filter(this.post.sections, s => !s.renderNode.isAttached()),
+      s => s.renderNode.scheduleForRemoval()
+    );
   }
 
   /*
@@ -571,8 +520,9 @@ class Editor {
     section.renderNode.markDirty();
   }
 
-  reparseSection(section) {
-    this._parser.reparseSection(section, this._renderTree);
+  _reparseCurrentSection() {
+    const {headSection:currentSection } = this.cursor.offsets;
+    this._parser.reparseSection(currentSection, this._renderTree);
   }
 
   serialize() {
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -1,6 +1,8 @@
 import { MARKUP_SECTION_TYPE } from '../models/markup-section';
+import { LIST_ITEM_TYPE } from '../models/list-item';
 import {
-  filter
+  filter,
+  compact
 } from '../utils/array-utils';
 
 import { DIRECTION } from '../utils/key';
@@ -9,9 +11,18 @@ function isMarkupSection(section) {
   return section.type === MARKUP_SECTION_TYPE;
 }
 
+function isListItem(section) {
+  return section.type === LIST_ITEM_TYPE;
+}
+
+function isBlankAndListItem(section) {
+  return isListItem(section) && section.isBlank;
+}
+
 class PostEditor {
   constructor(editor) {
     this.editor = editor;
+    this.builder = this.editor.builder;
     this._completionWorkQueue = [];
     this._didRerender = false;
     this._didUpdate = false;
@@ -99,12 +110,11 @@ class PostEditor {
   }
 
   _coalesceMarkers(section) {
-    let {builder} = this.editor;
     filter(section.markers, m => m.isEmpty).forEach(m => {
       this.removeMarker(m);
     });
     if (section.markers.isEmpty) {
-      section.markers.append(builder.createBlankMarker());
+      section.markers.append(this.builder.createBlankMarker());
       section.renderNode.markDirty();
     }
   }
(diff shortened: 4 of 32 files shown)


bantic:
@mixonic Could use review. This is ready to go now. After review I can update the standalone renderers to handle lists

tests/test-helpers.js
```diff
@@ -7,7 +7,24 @@ import skipInPhantom from './helpers/skip-in-phantom';
 import MobiledocHelpers from './helpers/mobiledoc';
 import PostAbstract from './helpers/post-abstract';
 
-const { test } = QUnit;
+const { test:qunitTest, module } = QUnit;
+
+QUnit.config.urlConfig.push({
+  id: 'debugTest',
+  label: 'Debug Test'
+});
```
bantic: this is a little odd, perhaps, but it greatly improves debugging speed for acceptance tests. This adds a checkbox 'DebugTest' in the QUnit toolbar. Clicking it causes the page to reload and drop into the debugger right before starting the test. The flow I've been using is:

  * run all tests (with dev tools open), click "rerun test" on any failing acceptance test
  * click the "Debug Test" checkbox -> page reloads and the debugger activates
  * step into the `originalCallback` test function and step through the lines in the test

I've found this to be a lot quicker than finding the test code in my text editor and reading through the steps to figure out what it is testing and what should be happening.

The caveat is that you need to `import {test, module} from Helpers` in the acceptance test (instead of `import {test, module} from QUnit`)

.jshintrc
```diff
@@ -31,7 +31,7 @@
   "undef"         : true,     // Require all non-global variables be declared before they are used.
   "unused"        : true,     // Warn when variables are created but not used.
   "trailing"      : true,     // Prohibit trailing whitespaces.
-  "es3"           : true,     // Prohibit trailing commas for old IE 
+  "es3"           : false,     // Prohibit trailing commas for old IE
```
mixonic: Should the comment change here as well?

Not crazy about leaking in commas. Harmless I suppose.

src/js/commands/image.js
```diff
@@ -13,6 +13,8 @@ export default class ImageCommand extends Command {
     let beforeSection = headMarker.section;
     let afterSection = beforeSection.next;
     let section = this.editor.builder.createCardSection('image');
+    // the postEditor needs to know this section's parent in order to `insertSectionBefore`
+    section.parent = beforeSection.parent;
```
mixonic: This seems wrong- adoption to a list should decide the parent. More comments below.

src/js/editor/post.js
```diff
@@ -482,8 +528,8 @@ class PostEditor {
    * @public
    */
   insertSectionBefore(section, beforeSection) {
-    this.editor.post.sections.insertBefore(section, beforeSection);
-    this.editor.post.renderNode.markDirty();
+    section.parent.sections.insertBefore(section, beforeSection);
```
mixonic: This should use `beforeSection.parent` instead of `section.parent`. The `section` may already have a parent, but it would not be parent you want to insert into.

src/js/editor/editor.js
```diff
@@ -446,78 +446,27 @@ class Editor {
...
     this.rerender();
     this.trigger('update');
 
-    if (resetCursor) {
-      this.cursor.moveToSection(headSection, headSectionOffset);
-    }
+    this.cursor.moveToSection(headSection, headSectionOffset);
+  }
+
+  _removeDetachedSections() {
```
mixonic: Seems less efficient than the old check. Ideally there would be a way to have encapsulation of logic an an efficient check.

src/js/editor/editor.js
```diff
@@ -571,8 +520,9 @@ class Editor {
     section.renderNode.markDirty();
   }
 
-  reparseSection(section) {
-    this._parser.reparseSection(section, this._renderTree);
+  _reparseCurentSection() {
+    const {headSection:currentSection } = this.cursor.offsets;
```
mixonic: We should really be passing something in to reparse. A parser should not know about the cursor position.

src/js/editor/post.js
```diff
@@ -9,9 +11,24 @@ function isMarkupSection(section) {
...
+function isBlankAndLastListItem(section) {
+  return isListItem(section) &&
+    section.isBlank &&
+    section.parent.sections.tail.isBlank; // existing tail is blank (FIXME)
+}
+
 class PostEditor {
   constructor(editor) {
     this.editor = editor;
+    this.builder = this.editor.builder;
```
mixonic: Should update the docs to use `postEditor.builder` if we provide this. Seems better than

```
editor.run((postEditor) = {
  postEditor.insertSectionBefore(editor.builder.createMarkupSection('p'), editor.post.sections.head);
});
```

My main justification for hesitancy is that the builder is non-mutating. It just returns new things. But leaving it on the editor for public consumers encourages them to much with the post outside a `run` block, so I am :+1: on putting it on the `postEditor`.

src/js/editor/post.js
```diff
@@ -9,9 +11,24 @@ function isMarkupSection(section) {
   return section.type === MARKUP_SECTION_TYPE;
 }
 
+function isTextEditableSection(section) {
```
mixonic: Are the semantics here still meaningful, or perhaps this was intended to have other complexity?

src/js/models/list-item.js
```diff
@@ -0,0 +1,51 @@
+import Section from './markup-section';
+import LinkedList from '../utils/linked-list';
+
+export const LIST_ITEM_TYPE = 'list-item';
+
+export default class ListItem extends Section {
+  constructor(tagName, markers=[]) {
+    super(tagName);
```
mixonic: Why not pass `markers` as well and share the logic for the linked list creation and append?

src/js/models/list-item.js
```diff
@@ -0,0 +1,51 @@
...
+    // FIXME splitAtMarker for listItem duplicates its superclass logic
+    let [beforeSection, afterSection] = [
+      this.builder.createListItem(),
+      createNewSection ? this.builder.createMarkupSection('p') : this.builder.createListItem()
+    ];
+
+    // for the postEditor to know how to `insertBefore` these sections,
+    // they must have a parent specified
+    beforeSection.parent = this.parent;
+    afterSection.parent = createNewSection ? this.parent.parent : // the post
```
mixonic: Should be able to go away as well.

src/js/models/markup-section.js
```diff
@@ -100,6 +94,20 @@ export default class Section extends LinkedItem {
...
+  splitAtMarker(marker, offset=0) {
+    let [beforeSection, afterSection] = [
+      this.builder.createMarkupSection(this.tagName, []),
+      this.builder.createMarkupSection(this.tagName, [])
+    ];
+
+    // for the postEditor to know how to `insertBefore` these sections,
+    // they must have a parent specified
+    beforeSection.parent = this.parent;
+    afterSection.parent = this.parent;
```
mixonic: Also here.

src/js/models/list-section.js
```diff
@@ -0,0 +1,68 @@
...
+
+export const LIST_SECTION_TYPE = 'list-section';
+
+export default class ListSection extends Section {
+  constructor(tagName, items=[]) {
+    super(tagName);
+    this.type = LIST_SECTION_TYPE;
+
+    // remove the inherited `markers` because they do nothing on a ListSection but confuse
+    this.markers = undefined;
```
mixonic: mmhm. Right. I almost think there is an alternative strategy of having `ListItem` and `MarkupSection` extend from `Markerable` instead of having `ListSection` inherit from `MarkupSection`. The bulk of the shared behaviors are probably around markers.

src/js/models/list-section.js
```diff
@@ -0,0 +1,68 @@
...
+    const prevItem = listItem.prev,
+          nextItem = listItem.next;
+    const listSection = this;
+    const parent = listSection.parent; // post
+
+    let prevListSection, nextListSection, newSection;
+
+    const createNewListSection = () => {
+      const section = this.builder.createListSection(this.tagName);
+      section.parent = this.parent;
```
mixonic: And here.

src/js/models/list-section.js
```diff
@@ -0,0 +1,68 @@
...
+      return section;
+    };
+
+    newSection = this.builder.createMarkupSection('p');
+    newSection.parent = parent;
```
mixonic: Here.

src/js/parsers/mobiledoc.js
```diff
@@ -46,15 +50,18 @@ export default class MobiledocParser {
...
         break;
-      case IMAGE_SECTION_TYPE:
+      case MOBILEDOC_IMAGE_SECTION_TYPE:
         this.parseImageSection(section, post);
         break;
-      case CARD_SECTION_TYPE:
+      case MOBILEDOC_CARD_SECTION_TYPE:
         this.parseCardSection(section, post);
         break;
+      case MOBILEDOC_LIST_SECTION_TYPE:
```
mixonic: un-certain about leaking `MOBILEDOC` onto the type names. There is nothing in the post abstract that is mobiledoc specific. `POST_` or `ABSTRACT_` or `AST_`?

mixonic: Ah sorry these *are* mobiledoc types! Excellent :-)

src/js/parsers/post.js
```diff
@@ -34,51 +29,71 @@ export default class PostParser {
...
-      const text = sanitizeText(textNode.textContent);
-      let markups = collectMarkups(textNode, sectionElement);
+    walkTextNodes(element, (textNode) => {
+      const text = textNode.textContent;
+      let markups = this.collectMarkups(textNode, element);
 
       let marker;
 
-      let renderNode = renderTree.getElementRenderNode(textNode);
+      let renderNode = renderTree.elements.get(textNode);
```
mixonic: Should likely remain `getElementRenderNode`

src/js/renderers/editor-dom.js
```diff
@@ -104,6 +111,37 @@ function renderMarker(marker, element, previousRenderNode) {
   return textNode;
 }
 
+function attachRenderNodeElementToDOM(renderNode, element, originalElement) {
```
mixonic: :tada:

src/js/renderers/editor-dom.js
```diff
@@ -258,28 +307,31 @@ let destroyHooks = {
...
   }
 };
 
 // removes children from parentNode that are scheduled for removal
 function removeChildren(parentNode) {
   let child = parentNode.childNodes.head;
+  let nextChild, method;
   while (child) {
-    let nextChild = child.next;
+    nextChild = child.next;
```
mixonic: Seems good, though also suspect is anything in the destroy hooks themselves is removing a renderNode. Render nodes should only be removed by the renderer itself (here on line 335, after a next child was read).

tests/test-helpers.js
mixonic: :+1: seems cool, eager to give it a try

.jshintrc
bantic: the comment can stay as it is; those are stock comments from default jshintrc and they all say what happens if the param is true (there are others that are false that aren't listed here)
I'm not psyched about relaxing this restriction either, but I figure it's a mechanical change to revert back later, if necessary (just flip this switch and see which jshint tests fail), and I was wasting a lot of cycles fixing dangling commas when changing around tests and other things.

src/js/editor/post.js
bantic: :+1: yeah, good catch, thanks.

src/js/editor/editor.js
bantic: The idea here was to push some complexity down from `reparse` to clarify what it does ("reparse current section", then "remove detached sections"). The specificity of this method also helps to clarify that we only ever need to reparse the section that the cursor is in, which wasn't obvious (to me) before I made this change.

I can change it to accept a section to reparse — perhaps a default arg?
I also notice there is a typo in the method name that I will fix.

src/js/editor/post.js
bantic: Yeah, this is a little confusing, I agree. I was expecting this to deal with the other text-ish section types (list section, list item) so opted for a more semantic name (rather than the rigid `isMarkupSection`), but later refactorings changed how those are dealt with and this ended up not needing any complexity. I could change the other parts of the post editor to use isMarkupSection instead.

src/js/models/list-item.js
bantic: Yeah, that would work. Originally the constructors for these two were not so similar, but later refactorings removed the differences. I'll change this.

src/js/models/list-section.js
bantic: Yeah, agreed. I was hoping to see more of an opportunity to build these sections up through composition rather than more granular inheritance levels, but using a `Markerable` section type does seem worthwhile.

src/js/renderers/editor-dom.js
bantic: I am not sure if any logic changed here from what it was before; I rearranged some variable declarations in order to throw a more meaningful error ("cannot destroy 'list-item'", e.g.) below on line 332

src/js/editor/editor.js
mixonic: I'm OK with punting the removal of cursors, since there was already something in the parser re: cursors before. Let's just keep the goalpost in mind for now.

.jshintrc
mixonic: It seems like this should just be fixed in this PR then :-/ Leaking it to master seems to provide very little long/mid-term value, though I 100% understand the short-term aid during dev.

src/js/editor/post.js
mixonic: Should probably just remove it if the value of the abstraction has been diminished/removed.

.jshintrc
bantic: fixed

src/js/editor/editor.js
bantic: oh — in case it isn't obvious, this method is on the editor, not the post parser. The post parser itself has no knowledge of cursors.

mixonic:
sweeeeeeeeetttttt

PR #95, opened 2015-08-27
Introduce lifecycle hooks

mixonic:
* didUpdatePost, which gets passed the postEditor
* willRender, before render after post changes are complete
* didRender, after render when DOM and post are in sync

TODO:

* [x] Docs

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -74,6 +74,13 @@ const defaults = {
   html: null
 };
 
+function runCallbacks(callbacks, args) {
+  let i;
+  for (i=0;i<callbacks.length;i++) {
+    callbacks[i].apply(null, args);
+  }
+}
+
 function bindContentEditableTypingListeners(editor) {
   // On 'PASTE' sanitize and insert
   editor.addEventListener(editor.element, 'paste', function(e) {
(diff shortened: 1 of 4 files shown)


mixonic:
@bantic this is good to go.

src/js/editor/editor.js
```diff
@@ -319,6 +332,8 @@ class Editor {
       showForTag: 'a'
     }));
 
+    // Stub out a post editing session for initial render
+    this.run(() => {});
```
bantic: Why is the empty `run` there? It's not clear to me from that comment.

mixonic: Will elaborate, poor wording there.

mixonic:
:tada:

src/js/editor/editor.js
```diff
@@ -319,6 +332,9 @@ class Editor {
       showForTag: 'a'
     }));
 
+    // A call to `run` will trigger the didUpdatePostCallbacks hooks with a
+    // postEditor.
+    this.run(() => {});
```
bantic: Maybe this should be `this.runCallbacks()` or something? This looks like the sort of no-op-y code that will get deleted by someone in the future.

PR #100, opened 2015-08-27
Drop all blank markers, section renders br

mixonic:
Before this patch, Content-Kit would insert blank markers in any empty section to ensure a cursor could be placed there. This changes that to be a section responsibility. When there are no child markers, a markup section will create a br.

This means throughout the codebase sections and section offset should be used to manage editing actions. There may be no marker at a given cursor position, only using the offset in a section is reliable.

Also ensures correct behavior when deleting two sections in their entirety.

### DIFF
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -221,14 +221,22 @@ class PostEditor {
         return this._deleteForwardFrom(result);
       }
     } else {
-      if (direction === DIRECTION.BACKWARD && section.prev) {
-        let prevSection = section.prev;
-        prevSection.join(section);
-        prevSection.renderNode.markDirty();
-        this.removeSection(section);
-        this.scheduleRerender();
-        this.scheduleDidUpdate();
-        return { currentSection: prevSection, currentOffset: prevSection.text.length };
+      if (direction === DIRECTION.BACKWARD) {
+        if (isMarkupSection(section) && section.prev) {
+          let prevSection = section.prev;
+          prevSection.join(section);
+          prevSection.renderNode.markDirty();
+          this.removeSection(section);
+          this.scheduleRerender();
+          this.scheduleDidUpdate();
+          return { currentSection: prevSection, currentOffset: prevSection.text.length };
+        } else if (isListItem(section)) {
+          this.scheduleRerender();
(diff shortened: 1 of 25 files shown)


src/js/editor/post.js
```diff
@@ -90,33 +91,90 @@ class PostEditor {
...
+      marker = nextMarker;
+    }
   }
 
   _coalesceMarkers(section) {
-    let {builder} = this.editor;
-    filter(section.markers, m => m.isEmpty).forEach(m => {
-      this.removeMarker(m);
+    filter(section.markers, m => m.isEmpty).forEach(marker => {
+      if (marker.isEmpty) {
```
bantic: is this a redundant check?

mixonic: yes, thanks!

mixonic:
Pushed a successful rebase on master, though there are failing tests for lists. I presume their elements must be registered into the elements map, then many things should fall into place.

mixonic:
This afternoon we investigated changing the list section and items to work with this no-more-blank-marker-branch. We approached it by attempting to have offsets always be top-level section based. For example given this content:

> Howdy
> * Friend

And a selection including `dy\n* Fri`, we would expect these offsets:

```
{
  headSection: markup section with Howdy,
  headOffset: 3,
  tailSection: list section with Friend,
  tailOffset: 3
}
```

However we quickly realized that some list sections would then contain multiple list item sections that, having zero width if they are empty, would be un-addressable. For example:

> Howdy
> * 
> * Friend

Knowing the difference between the cursor being in the first list item and at the start of the second list item would not be possible.

With this reflection in mind, we believe offsets on cursors should always be pinned to post model that has markers in it. This is because items with markers in them, unlike anything else in the post model, are allowed to be empty.

Next steps are to introduce a "markerable" (poor name) mixin or super-class, update `offsets` to return those instead of sections, and make the various mutation APIs smart about how to handle the boundery conditions between different kinds of markerables.

For posterity, our WIP used for exploration: https://github.com/mixonic/content-kit-editor/tree/die-blank-markers-die-WIP

mixonic:
@bantic AWESOME!!!!! :rose:

PR #103, opened 2015-08-31
Bring back List Commands

bantic:


src/js/commands/list.js
```diff
@@ -1,40 +1,36 @@
...
+    const { head: {section:currentSection} } = cursor.offsets;
+
+    const listItem = editor.run(postEditor => {
+      const { builder } = postEditor;
+      const tagName = this.tag;
+      const listSection = builder.createListSection(tagName);
+      const listItem = builder.createListItem();
+      listSection.items.append(listItem);
+
+      postEditor.replaceSection(currentSection, listSection);
```
mixonic: Amaze.

src/js/models/_markerable.js
```diff
@@ -35,6 +35,31 @@ export default class Markerable extends LinkedItem {
...
+   * @param {Marker}
+   * @param {Number} markerOffset The offset relative to the start of the marker
+   *
+   * @return {Number} The offset relative to the start of this section
+   */
+  offsetOfMarker(marker, markerOffset) {
+    if (marker.section !== this) {
+      throw new Error(`Cannot get offsetOfMarker for marker that is not child of this`);
+    }
+    // FIXME it is possible, when we get a cursor position before having finished reparsing,
```
mixonic: There must be a nuance here. I'm not sure when we are asking for the offset of a marker, but against dirty or out of sync post ast. It seems bad to do that, I agree, but the problem seems to lie with asking for cursor position in a marker (beyond just the section) before parsing. Hm.

mixonic:
Seems pretty great at a glance.

PR #105, opened 2015-09-01
Add #remove env hook to cards

### LINKED ISSUE #104: Card#remove lifecycle hook
There needs to be a way to have a card remove itself from the Post abstract. This should be exposed as a hook that the card can call that will tear it down, remove it from the abstract, and remove its rendered element from DOM.

`#remove` seems like an appropriate name for this.

cc @dtetto

bantic:
fixes #104

cc @dtetto


bantic:
@dtetto Another possible name for this hook would be `destroy` — Any thoughts either way?

PR #110, opened 2015-09-03
Implement text expansions

### LINKED ISSUE #87: text expansion: "* " -> unordered list item, "1. " -> ordered list item
See also #86

bantic:
* Trigger is "space"
  * "*" -> ul>li
  * "1" and "1." -> ol>li
  * "##" -> h2
  * "###" -> h3

Adds `editor#registerExpansion` which takes an expansion.
An expansion has these keys:
  * `trigger` the keycode of the triggering key event. All of the existing ones use 32 ("space")
  * `text` the text that must match to the left of the cursor. Currently this only looks for text matching from the beginning of the section to the cursor. In the future it wouldn't be hard to modify it to handle in-context text, so you could bold text by adding "**" on either side of a word, perhaps
  * `run` a method that will get passed the `editor` instance when this expansion is triggered

fixes #87

gpy

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -43,6 +43,11 @@ import mixin from '../utils/mixin';
 import EventListenerMixin from '../utils/event-listener';
 import Cursor from '../utils/cursor';
 import PostNodeBuilder from '../models/post-node-builder';
+import {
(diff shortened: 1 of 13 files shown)


bantic:
cc @ErisDS This was something you had asked about a while back. It's finally landed!

PR #113, opened 2015-09-03
Cleanup listeners

bantic:
Simplify and DRY the event-listening code so that it's easier to turn on/off editability by squashing events in `handleEvent`


bantic:
Should also make it easier to deal with any difficulties presented by using multiple editors at once.

PR #147, opened 2015-09-17
Copy Paste

### LINKED ISSUE #111: Handle copy/pasting properly
When pasting text into an editor, we should try to add the pasted content to the existing post abstract, matching as closely as possible.

We can also attempt to serialize sections of mobiledoc when copying by overriding the `copy` event and either setting a custom mime type like "text/mobiledoc" (works on Safari and Chrome) or hijacking the "text/html" mime type (works on Firefox), and prioritize using that if it is found. This makes pasting easier/reliable when working within the same document (which seems like a common use case), and allows cross-window and cross-browser copy/pasting of mobiledoc.

Here's an example codepen showing overridden copied values in action — http://codepen.io/bantic/pen/dYPWea?editors=101

bantic:
*ready to merge*

This adds support for copy-pasting content from one Content Kit to another, and from Google Docs to Content-Kit.

  * Rename PostParser -> DOMParser, remove old DOMParser
  * Add HTMLParser (parses from HTML to PostAbstract)
  * Add many fixtures with example HTML from copying in a google doc document
  * Add postEditor#insertPost. It inserts the sections from a post into the current editor's post at the passed Position, with some logic for merging/inserting into the section with the cursor in it. It returns a Position used to position the cursor after inserting content
  * Only allow certain attributes from parsed content (currently only "href" and "rel" are acceptable attributes. These are defined in Markup's `VALID_ATTRIBUTES`)

  * [x] support for pasting multi-section copy data
  * [x] support for pasting unstructured HTML
  * [x] lock down allowed element attributes when parsing HTML

For a separate PR:

  * [x] support for copying non-markerable sections like cards (and list sections)
  * [ ] parse lists properly from HTML. They currently cause a MarkupSection with tagName "li" to be created, but we want them to cause a ListSection and ListItem(s) to be created instead. (#183)

For later:

  * [ ] Add the ability to handle more exotic pasted data, such as images.

fixes #111


bantic:
Many tests added for parsing google-docs-style copy/pasted HTML

src/js/parsers/section.js
```diff
@@ -6,10 +6,17 @@ import {
   VALID_MARKUP_SECTION_TAGNAMES
 } from 'content-kit-editor/models/markup-section';
 
+function popMany(arr, count) {
```
mixonic: seems like good use for `splice`. These should be the same:

```js
popMany(array, 5);
array.splice(-5, 5, []);
```

mixonic:
Lookin' pretty sweet @bantic :-) :-) :-) :-) :-)

src/js/parsers/section.js
bantic: :+1: it's actually `array.splice(-length, +length)` (no `[]`), but yeah. Changing this. I did that because I can never remember the arguments to splice

bantic:
@mixonic Ready for :eyeglasses: when you can.

copypaste

bantic:
If you're curious to see what the HTML result of copy/pasting something from Google docs looks like, you can use [this tool](http://codepen.io/bantic/pen/wKpOqE) to see it.

src/js/editor/editor.js
```diff
@@ -640,8 +642,27 @@ class Editor {
     return false;
   }
 
+  handleCopy(event) {
+    const { clipboardData } = event;
+    const range = this.cursor.offsets;
+    const mobiledoc = this.post.cloneRange(range);
+    const html = `<div data-mobiledoc='${JSON.stringify(mobiledoc)}'></div>`;
+    clipboardData.setData('text/html', html);
+    event.preventDefault();
```
mixonic: @bantic this logic means we cannot meaningfully copy *out* of Content-Kit to third parties?

bantic: Yeah.
The preventDefault is necessary for the text/html that we set on the above line to be preserved.
Rendering the HTML with the HTML Renderer and inserting it in the div above would preserve copy-to-outside-world functionality.

bantic: Or, honestly, just letting natural browser copy do its thing for now would work prety well. We'll reparse the HTML on the way back in rather than using the already-serialized mobiledoc, but that doesn't seem to be a problem. Eventually we will want to keep that serialized mobiledoc so that cards can be preserved and copy/pasted around.

mixonic: Aw man that would be cool. Opened a followup https://github.com/bustlelabs/content-kit-editor/issues/180

src/js/models/markup.js
```diff
@@ -10,27 +11,36 @@ export const VALID_MARKUP_TAGNAMES = [
   'li'
 ].map(normalizeTagName);
 
+export const VALID_ATTRIBUTES = [
```
mixonic: I see this is used as a filter during parsing, but should it also be used for an assertion here as `VALID_MARKUP_TAGNAMES` etc is? IMO yes.

src/js/models/post.js
```diff
@@ -175,4 +177,19 @@ export default class Post {
       }
     }
   }
+
+  cloneRange(range) {
+    const post = this.builder.createPost();
+    const { builder } = this;
+    this.walkMarkerableSections(range, section => {
```
mixonic: This seems like it would discard copying card sections?

src/js/editor/editor.js
mixonic: > Or, honestly, just letting natural browser copy do its thing for now would work prety well. We'll reparse the HTML on the way back in rather than using the already-serialized mobiledoc, but that doesn't seem to be a problem. Eventually we will want to keep that serialized mobiledoc so that cards can be preserved and copy/pasted around.

Ah, per another comment here I was surprised not to see cards being copied. I think both these things need followups if they are not in scope here. Copying mobiledocs cards should be pretty trivial.

src/js/models/markup.js
bantic: Yeah, I think so — and in fact, VALID_ATTRIBUTES should most likely be contextual. `href` isn't valid on a `<em>` tag, for instance. I'll move the attribute-filtering (as is) into Markup directly, for now.

src/js/editor/editor.js
bantic: Cool, after this merges I'll add a follow-up PR to address #180.

bantic:
needs to handle "cut" in addition to "copy" event

PR #154, opened 2015-09-21
Handle a section of tagName "pull-quote"

### LINKED ISSUE #153: add "pull-quote" section type
See https://github.com/bustlelabs/typeset.bustle.com/issues/12

bantic:
fixes #153


mixonic:
Should this have a corresponding change to the parser so `div.pull-quote` elements are parsed into the markup section type?

I don't expect there should be any in the DOM that are not backed by the abstract. The only case that comes to mind would be paste, but we've discussed having an alternate mobiledoc protocol there anyway. So the parsing case seems moot.

PR #156, opened 2015-09-21
Handle selecting sections when the passed array is empty

### LINKED ISSUE #155: editor#selectSections should be a no-op when there are no active sections

bantic:
fixes #155

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -287,9 +287,14 @@ class Editor {
     this.trigger('update');
   }
 
-  selectSections(sections) {
-    this.cursor.selectSections(sections);
-    this.reportSelection();
+  selectSections(sections=[]) {
+    if (sections.length) {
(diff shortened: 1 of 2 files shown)


src/js/editor/editor.js
```diff
@@ -287,9 +287,14 @@ class Editor {
...
-  selectSections(sections) {
-    this.cursor.selectSections(sections);
-    this.reportSelection();
+  selectSections(sections=[]) {
+    if (sections.length) {
+      this.cursor.selectSections(sections);
+      this.reportSelection();
+    } else {
+      this.cursor.clearSelection();
+      this.reportNoSelection();
```
mixonic: Should reporting of selection state be something the cursor does? It can fire a callback and that can trigger the complex callbacks on the editor

mixonic: I see, reporting code is already leaked through the editor elsewhere.

PR #158, opened 2015-09-21
add `cursorDidChange` lifecycle hook to editor, called when cursor position changes

### LINKED ISSUE #157: add `cursorChanged` event
This event will be useful in building a UI for content-kit that can react when the user moves the cursor through different markups, so that the UI can highlight a toolbar button corresponding to the current text under the cursor, for instance.

bantic:
The editor runs these callbacks after the selection changes as result of a mouseup or keyup event.
fixes #157

As written now, this doesn't guarantee that the callbacks run once-and-only-once when the cursor moves, and also doesn't guarantee that the cursor did actually change position.

As implemented, though, this is enough for a UI element (like a toolbar button)
to stay up to date as the user selects, clicks, or arrows through the text.
Improving cursor position tracking to not fire on, say, multiple clicks in the 
exact same location would be nice, but does not seem like a critical requirement
for a first iteration of this event.

cc @mixonic

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ The available lifecycle hooks are:
    the DOM is updated.
 * `editor.didRender()` - After the DOM has been updated to match the
   edited post.
+* `editor.cursorDidChange()` - When the cursor (or selection) changes as a result of arrow-key
+  movement or clicking in the document.
 
 ### Programmatic Post Editing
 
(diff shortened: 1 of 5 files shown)


mixonic:
Is this intended to be the public API? If so, can it use the hooks API and be documented as [here](https://github.com/bustlelabs/content-kit-editor#editor-lifecycle-hooks)? I was moving away from the trigger/on API for public interface.

bantic:
@mixonic Sure, we could put it through the lifecycle hooks. Which one do you think makes the most sense?

mixonic:
Awesome refactor!! :-D

PR #159, opened 2015-09-22
Add #detectMarkupInRange to editor

mixonic:
Add a public API for detecting markup in a range. This is used by Ember to see if a link is present in the current selection.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -34,6 +34,8 @@ import LifecycleCallbacksMixin from '../utils/lifecycle-callbacks';
 
 export const EDITOR_ELEMENT_CLASS_NAME = 'ck-editor';
 
+import { detect } from '../utils/array-utils';
+
 const defaults = {
   placeholder: 'Write here...',
   spellcheck: true,
@@ -227,7 +229,7 @@ class Editor {
       this.cursor.moveToPosition(range.head);
     } else {
       const key = Key.fromEvent(event);
-      const nextPosition = this.run(postEditor => { 
+      const nextPosition = this.run(postEditor => {
         return postEditor.deleteFrom(range.head, key.direction);
       });
       this.cursor.moveToPosition(nextPosition);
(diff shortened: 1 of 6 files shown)


src/js/editor/post.js
```diff
@@ -602,20 +602,23 @@ class PostEditor {
...
-
-    const range = this.editor.cursor.offsets;
-    const hasMarkup = m => m.hasTag(markup.tagName);
-    const rangeHasMarkup = any(this.editor.markupsInSelection, hasMarkup);
-
-    if (rangeHasMarkup) {
-      this.removeMarkupFromRange(range, hasMarkup);
-    } else {
-      this.applyMarkupToRange(range, markup);
+    if (this.editor.cursor.hasSelection()) {
```
bantic: can this be refactored to not need to know if there is a selection?
`removeMarkupFromRange` and `applyMarkupToRange` should be safe to run on a collapsed range.

src/js/editor/editor.js
```diff
@@ -384,6 +386,13 @@ class Editor {
     return activeSections[activeSections.length - 1];
   }
 
+  detectRangeMarkup(range, markupTagName) {
```
bantic: perhaps `hasMarkupInRange` ? or `detectMarkupInRange`?

bantic:
looks good, a few comments. Perhaps at some point we should add markup-related methods should work by tagName+attributes rather than the underlying object, and symmetric methods like:

`editor#hasMarkupInRange(range, tagName, optionalAttributes)` -> boolean
`postEditor#removeMarkupFromRange(range, tagName, optionalAttributes)`
`postEditor#addMarkupToRange(range, tagName, optionalAttributes)`

src/js/editor/editor.js
mixonic: `hasMarkupInRange` implies a boolean, and this does actually return the first one found ala a `detect`

bantic: makes sense. I was confused because elsewhere the variable name for the result of `detectRangeMarkup` was being called `hasMarkup`

src/js/editor/post.js
mixonic: I can- but I'm not sure this will ever run without a selection. Without a selection we would do something else like run the stateful code.

And avoiding the `selectRange` when it is not needed (when it is not broken) seems ideal.

mixonic:
@bantic I've made several suggested changes, but held off from removing the `isCollapsed` check from `toggleMarkup`. There should be no reason to create a new selection if no work is being done. I did however write some tests to ensure the upstream methods used by `toggleMarkup` behave reasonably when given a blank range.

PR #160, opened 2015-09-23
Add postEditor#moveSectionUp, postEditor#moveSectionDown

bantic:
In order to move a section that is already rendered, we have to remove it
and put a copy of it in the right place. This PR also adds a `clone` method to
sections for that purpose.

This PR removes `builder#createBlankPost`. It previously created a post with
a single empty markup section because posts couldn't have 0 sections, but that
restriction was relaxed recently. Upstream consumers of content-kit were getting
confused by having single empty markup section when booting an editor.

cc @mixonic

### DIFF
--- a/src/js/models/card-node.js
+++ b/src/js/models/card-node.js
@@ -35,7 +35,8 @@ export default class CardNode {
         this.display();
       },
       cancel: () => this.display(),
-      remove: () => this.remove()
+      remove: () => this.remove(),
+      section: this.section
     };
   }
 
--- a/CARDS.md
+++ b/CARDS.md
@@ -66,7 +66,8 @@ var exampleCard = {
   The content for the card should be pushed on that array as a string.
 * `options` is the `cardOptions` argument passed to the editor or renderer.
 * `env` contains information about the running of this hook. It may contain
-  the following functions:
+  the following properties:
+  * `env.name` The name of this card
   * `env.save(payload)` will save a new payload for a card instance, then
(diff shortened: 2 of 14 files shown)


src/js/models/card-node.js
```diff
@@ -35,7 +35,8 @@ export default class CardNode {
         this.display();
       },
       cancel: () => this.display(),
-      remove: () => this.remove()
+      remove: () => this.remove(),
+      card: this
```
mixonic: `this` is the card node, not the instance of the card as a user sees it (`this.card[mode]` I believe). I'm not sure making the `CardNode` public API for card implementers is ideal.

mixonic: I see how this is being used.

Could this expose the abstract node instead of the `CardNode`? The abstracts are already public or at the least "intimate" API, and unless I've read incorrectly the only other thing needed in https://github.com/bustlelabs/typeset.bustle.com/pull/21/files#diff-04c1550a2081f24687e1031a3bae973fR13 is the `editor` itself, which could be passed directly into the component.

mixonic: https://github.com/bustlelabs/content-kit-editor/blob/master/CARDS.md should be updated as well.

bantic:
updated to add tests for all the properties that a card's `env` is expected to have, including `section` (the cardSection) instead of the `cardNode` itself. Updated the docs in CARDS.md

PR #161, opened 2015-09-23
Drop floating and sticky toolbar

mixonic:
The floating and sticky toolbars in Content-Kit are not going to survive until 1.0. This PR begins their removal.

TODO:

* [x] publish the Ember addon, and update the demo to use the Ember addon for UX


tests/acceptance/editor-selections-test.js
```diff
@@ -156,13 +159,14 @@ test('select text and apply markup multiple times', (assert) => {
...
-    Helpers.toolbar.clickButton(assert, 'bold');
+    let command = new BoldCommand(editor);
+    command.exec();
 
     Helpers.dom.selectText('fir', editorElement);
     Helpers.dom.triggerEvent(document, 'mouseup');
 
     setTimeout(() => {
-      Helpers.toolbar.clickButton(assert, 'bold');
+      command.exec();
```
bantic: there should no longer be a need to use `setTimeout` (or triggering "mouseup") in these tests.

tests/unit/commands/heading-test.js
```diff
@@ -0,0 +1,142 @@
...
+test('click heading button triggers update', (assert) => {
+  const done = assert.async();
+  const triggered = [];
+  const triggerFn = editor.trigger;
+  editor.trigger = (name, ...args) => {
+    triggered.push(name);
+    triggerFn.call(editor, name, ...args);
+  };
+
+  setTimeout(() => {
```
bantic: same as above, I do not think that there is async involved in this code path any longer

tests/unit/editor/editor-destroy-test.js
```diff
@@ -1,48 +0,0 @@
...
-    editor.render(editorElement);
-  },
-  afterEach() {
-    if (editor && !editor._isDestroyed) {
-      editor.destroy();
-    }
-  }
-});
-
-test('removes toolbar from DOM', (assert) => {
```
bantic: it may still be worth testing that the remaining views (such as the tooltip) added to the editor get destroyed, and that event listeners are unregistered, if that is not tested elsewhere.

PR #167, opened 2015-10-09
register arbitrary custom key commands

rlivsey:
registerKeyCommand currently requires the string format of the key and a modifier.

This adds the ability to register arbitrary commands by supplying a `check` function 
with the command which is passed the key event.

### DIFF
--- a/src/js/editor/key-commands.js
+++ b/src/js/editor/key-commands.js
@@ -47,13 +47,20 @@ export const DEFAULT_KEY_COMMANDS = [{
 }];
 
 export function validateKeyCommand(keyCommand) {
-  return !!keyCommand.modifier && !!keyCommand.str && !!keyCommand.run;
+  if (!keyCommand.run) {
+    return false;
+  }
+  return !!keyCommand.check || (!!keyCommand.modifier && !!keyCommand.str);
 }
 
 export function findKeyCommand(keyCommands, keyEvent) {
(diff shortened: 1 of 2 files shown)


bantic:
One of the travis runs failed due to a bower issue, I just kicked it to run it again.

src/js/editor/key-commands.js
```diff
@@ -47,13 +47,20 @@ export const DEFAULT_KEY_COMMANDS = [{
...
 }
 
 export function findKeyCommand(keyCommands, keyEvent) {
   const key = Key.fromEvent(keyEvent);
 
-  return detect(keyCommands, ({modifier, str}) => {
-    return key.hasModifier(modifier) && key.isChar(str);
+  return detect(keyCommands, ({modifier, str, check}) => {
+    if (!!check) {
+      return check(keyEvent);
```
rlivsey: I think this should take a `Key` and not the raw event. Then we can use things like `key.isEnter()` from inside `check`

rlivsey: Also as mentioned in https://github.com/bustlelabs/content-kit-editor/issues/168, this could/should also receive the `editor` so decisions could be made as to whether to handle the event based on the current editor state.

rlivsey:
Closing in preference to the outcome of #168 and registering string representations like "CMD+A" etc… instead.

PR #172, opened 2015-10-14
Key commands refactor

### LINKED ISSUE #168: RFC - Fully custom keyboard events
Building on #167 and discussions with @mixonic I thought I'd jot my thoughts down here before getting too far down the rabbit hole with more pull requests.

We're in the process of integrating ember-content-kit in our app and have a few use-cases around keyboard event handling which aren't currently easily supported. For example, overriding enter and preventing a new-line being inserted (eg for comment submission fields), custom behaviour on tab etc...

Pull request #167 goes someway towards supporting this to allow registering arbitrary key commands (previously a modifier key had to be held down & the key pressed be a printable character).

[Currently][1] all internal event handling on `handleKeyDown` happens before calling custom key commands, which means we can't currently override things like delete/enter. 

```js
handleKeydown(event) {
  const key = Key.fromEvent(event);

  if (key.isDelete()) {
    this.handleDeletion(event);
    event.preventDefault();
  } else if (key.isEnter()) {
    this.handleNewline(event);
  } else if (key.isPrintable()) {
    // ...
  }

  this.handleExpansion(event);
  
  // can't handle 'enter' here as it's already happened
  this.handleKeyCommand(event);
}
```

Also, in `handleKeyCommand` `event.preventDefault` is [always called][2] so there can be no conditional handling in the command itself.

## Handle key commands first

I propose moving `handleKeyCommand` to happen before the internal behaviour, and if a command is handled then return at that point:

```js
handleKeydown(event) {
  const key = Key.fromEvent(event);

  if (this.handleKeyCommand(event)) {
    // if we've handled it, don't do anything else?
    return;
  else if (key.isDelete()) {
    this.handleDeletion(event);
    event.preventDefault();
  } else if (key.isEnter()) {
    this.handleNewline(event);
  } else if (key.isPrintable()) {
    // ...
  }

  this.handleExpansion(event);
}
```

## Pass `editor` to key command `check`

#167 adds a `check` function to the key command, but this doesn't receive the editor so cannot conditionally perform the command based on the state of the editor itself.

Eg. to override enter if the editor has content:

```js
editor.registerKeyCommand({
  check(key, editor) {
    return key.isEnter() && !editor.post.isBlank();
  },
  run(editor) {
  }
})
```

## Implement internal key commands using registerKeyCommand

Once the above are done, it would be possible to implement `handleDeletion` and `handleNewLine` etc as commands themselves:

```js
editor.registerKeyCommand({
  check(key) {
    return key.isDelete();
  },
  run(editor) {
    // current deletion behaviour
  }
})
```

However, that raises the issue of what happens if you register a duplicate command (eg to override handleDeletion). 

## Allow prepending commands

Currently `registerKeyCommand` [pushes onto the end of the stack][3] so if you wanted to register a command to override built-in behaviour then it would never be called as the first matching command is run and subsequent ones aren't.

Options here could be to add a flag to `registerKeyCommand`, a named parameter, or a separate method:

```js
editor.registerKeyCommand(command, true) // true to unshift vs push
editor.registerKeyCommand(command, { prepend: true })
editor.prependKeyCommand(command);
```

If commands were named, then we could also do something like:

```js
editor.registerKeyCommand(command, { before: "delete" })
```

... but that starts getting complex and I don't think that's necessary.

[1]:  https://github.com/bustlelabs/content-kit-editor/blob/e64872aca5ee6e59c5a22fc4cd70b17084769a72/src/js/editor/editor.js#L580-L603
[2]: https://github.com/bustlelabs/content-kit-editor/blob/e64872aca5ee6e59c5a22fc4cd70b17084769a72/src/js/editor/editor.js#L608
[3]: https://github.com/bustlelabs/content-kit-editor/blob/e64872aca5ee6e59c5a22fc4cd70b17084769a72/src/js/editor/editor.js#L199

rlivsey:
This refactors key commands so that:

* special keys can be registered ("enter", "tab", "backspace" etc…)
* combo strings can be registered ("CMD+C" vs separate modifier)
* duplicate keys can be registered and fall back if they return `false` from `run`
* built-in commands can be overridden (eg register "enter" to override new-line behaviour)

This removes the requirement that all registered keys have a modifier, so that you can register just "tab" etc… we may want to only allow this for "special" keys so you still have to have a modifier for regular characters?

Registering a combo string simply transforms it to a modifier (if present) and code combination, so is completely backwards compatible.

This doesn't add any support for multiple modifiers ("CTRL+CMD+X"), or for sequences ("up up down down left right left right B A").

Closes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,14 +124,14 @@ For more details on the API of `postEditor`, see the [API documentation](https:/
 For more details on the API for the builder, required to create new sections
 and markers, see the [builder API](https://github.com/bustlelabs/content-kit-editor/blob/master/src/js/models/post-node-builder.js).
 
-### Configuring hot keys and text expansions
+### Configuring hot keys
 
 Content-Kit allows configuring hot keys and text expansions. For instance, the
 hot-key command-B to make selected text bold, is registered internally as:
+
 ```javascript
 const boldKeyCommand = {
-  modifier: META,
-  str: 'B',
+  str: 'META+B',
   run(editor) {
     editor.run(postEditor => postEditor.toggleMarkup('strong'));
   }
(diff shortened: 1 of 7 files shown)


bantic:
This looks great to me. @rlivsey thanks for putting this together! @mixonic can you weigh in?

mixonic:
@rlivsey this looks great. Can I ask you to add something to the README as well?

rlivsey:
@mixonic done, hope that makes sense!

bantic:
@rlivsey thanks again!

PR #173, opened 2015-10-14
Add `editor#editCard` and `editor#displayCard` to change a card's render mode before/after render

bantic:
Cards normally render into "display" mode by default. Calling
`editor#editCard` before the card is rendered will change it so that the
card gets rendered in edit mode initially. If the card has already been
rendered, it switches it to edit mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,10 @@ editor.render(element);
   current post with their cursor. Programmatic edits are still allowed.
 * `editor.enableEditing()` - allow the user to make direct edits directly
   to a post's text.
+* `editor.editCard(cardSection)` - change the card to its edit mode (will change
+  immediately if the card is already rendered, or will ensure that when the card
+  does get rendered it will be rendered in the "edit" state initially)
+* `editor.displayCard(cardSection)` - same as `editCard` except in display mode.
 
 ### Editor Lifecycle Hooks
 
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -30,6 +30,7 @@ import {
 } from './key-commands';
 import { capitalize } from '../utils/string-utils';
 import LifecycleCallbacksMixin from '../utils/lifecycle-callbacks';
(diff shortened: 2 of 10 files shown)


tests/unit/editor/card-lifecycle-test.js
```diff
@@ -93,6 +93,44 @@ test('rendering a mobiledoc for editing calls #unknownCardHandler when it encoun
...
+
+  editor = new Editor({ mobiledoc, cards });
+  editor.render(editorElement);
+
+  // must add the card via `run` because the initialMode property would
+  // be lost (if it were set) when serializing the mobiledoc above
+  editor.run(postEditor => {
+    let { builder } = postEditor;
+    let card = builder.createCardSection('my-card');
+    card.initialMode = 'edit';
```
mixonic: Could the mode perhaps be part of the args to `createCardSection`?

bantic: yeah, definitely. I was toying with it as a third argument:
`builder.createCardSection(cardName, payload={}, initialMode='display')`

That seems better, right?

mixonic: should definitely not be on the payload, so yeah seems good.

mixonic:
We discussed a further iteration of this: The editor should expose a method to toggle a card in and out of `edit`. Something like `editor.editCard(cardSection);` and `editor.cancelEditingCard(cardSection);`. These can access the cardNode off the cardSection and move the card between edit and display modes.

IMO the followup is not necessary: If a cardSection has not been rendered it will have no cardNode. In these cases the AST should store a hint that the later-instantiated cardNode can use to decide which was the card will be rendered.

bantic:
refactoring to add the `editCard` api as described ^^

bantic:
@mixonic refactored as described above. I'll merge this in soon to support the ember-content-kit PR. Lmk if you have feedback.

mixonic:
Very cool.

Issue #96, opened 2015-08-27
control-K clears section, causes cursor placement errors


bantic:
On Mac, Chrome:
Hitting control-K at the beginning of a section clears the section and an error appears in the console: "cannot move cursor to section without a marker".

bantic:
Better semantic cursor position detection (via #115) has fixed this problem.

mixonic:
Reopening this, it should be implemented by a keyboard command in content-kit

bantic:
A key command for this was [added in 7774864](https://github.com/bustlelabs/content-kit-editor/blob/7774864980b6944edd755346044ff13f730fa299/src/js/editor/key-commands.js#L48-L57)

PR #179, opened 2015-10-22
execCommands for bold, italic without selection

mixonic:
Currently, Content-Kit still parses the DOM after each keystroke. We are pretty good at it (text and DOM nodes from parsed section are re-used instead of being rerendered) however it does mean we cannot easily create our own stateful markup system for bold and italic. Specifically, for when bold or italic are toggled without a selection.

This patch restores a native behavior for contentEditables: Toggling italic or bold without a selection will just call the proper execCommand. This then applied to the next input.

The compromise is that execCommands vary from implementation to implementation, for example on Chrome they output a `b` tag, however on IE a `strong` tag is generated. See the following link for more details:

http://help.dottoro.com/ljcvtcaw.php

Here we normalize the `b` tag during section parse (there are two DOM parsers, this is the "safe" DOM one) into a `strong` markup. This means buttons toggling `strong` continue to work and the `b` tag is not
persisted into the mobiledoc. However, the `b` tag stays in DOM.

Additionally:

* Cleans up setup/teardown for some tests
* Use `ok` and equality checks to avoid QUnits terrible diffing system for   displayed failures of `equal`
* During reparse, we were resetting the cursor position. This was related to ctrl-k on OSX, which clears to the end of the line. Here that functionality is re-implemented as a key command. https://github.com/bustlelabs/content-kit-editor/issues/96

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -305,13 +305,6 @@ class Editor {
   }
 
   reparse() {
-    let { headSection, headSectionOffset } = this.cursor.offsets;
-    if (headSectionOffset === 0) {
-      // FIXME if the offset is 0, the user is typing the first character
-      // in an empty section, so we need to move the cursor 1 letter forward
-      headSectionOffset = 1;
-    }
-
     this._reparseCurrentSection();
     this._removeDetachedSections();
 
@@ -320,8 +313,6 @@ class Editor {
     this.run(() => {});
     this.rerender();
     this.trigger('update');
-
-    this.cursor.moveToSection(headSection, headSectionOffset);
   }
 
   // FIXME this should be able to be removed now -- if any sections are detached,
@@ -595,13 +586,12 @@ class Editor {
   }
 
   handleKeydown(event) {
-    if (!this.isEditable) { return; }
-    if (this.post.isBlank) {
-      this._insertEmptyMarkupSectionAtCursor();
+    if (!this.isEditable || this.handleKeyCommand(event)) {
+      return;
     }
 
-    if (this.handleKeyCommand(event)) {
-      return;
+    if (this.post.isBlank) {
+      this._insertEmptyMarkupSectionAtCursor();
     }
 
     const key = Key.fromEvent(event);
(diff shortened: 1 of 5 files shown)


tests/acceptance/editor-sections-test.js
```diff
@@ -88,14 +88,18 @@ const mobileDocWithNoCharacter = {
 
 module('Acceptance: Editor sections', {
   beforeEach() {
-    fixture = document.getElementById('qunit-fixture');
-    editorElement = document.createElement('div');
-    editorElement.setAttribute('id', 'editor');
-    fixture.appendChild(editorElement);
+    fixture = $('#qunit-fixture');
+    fixture.empty();
```
bantic: This should be unnecessary. It's the default behavior for QUnit.

tests/acceptance/editor-sections-test.js
```diff
@@ -88,14 +88,18 @@ const mobileDocWithNoCharacter = {
...
+    fixture.append(editorElement);
   },
 
   afterEach() {
-    if (editor) { editor.destroy(); }
+    if (editor) {
+      editor.destroy();
+      editor = null;
+    }
+    fixture.empty();
```
bantic: As mentioned above, this can be removed.

src/js/editor/editor.js
```diff
@@ -305,13 +305,6 @@ class Editor {
   }
 
   reparse() {
-    let { headSection, headSectionOffset } = this.cursor.offsets;
-    if (headSectionOffset === 0) {
-      // FIXME if the offset is 0, the user is typing the first character
-      // in an empty section, so we need to move the cursor 1 letter forward
-      headSectionOffset = 1;
-    }
```
bantic: The only thing that I know of that this was still critical for is handling the OS X "Document shortcut" Control-K (see [apple's docs](https://support.apple.com/en-us/HT201236) for more info) that clears to end of line. Without manually replacing the cursor, it goes shooting off somewhere crazy.

We should just reopen #96 and add a key command for control-K, though.

bantic:
This is really great. Happy to see this landing in Content Kit finally.

src/js/editor/editor.js
mixonic: Cool, reimplemented as a key command here.

mixonic:
Awesome, getting in master to test some more.

PR #188, opened 2015-10-23
Parse ul and ols correctly

### LINKED ISSUE #183: parse pasted HTML that includes LI items properly
after #147 is merged, pasting HTML that has list items will end up being parsed (incorrectly) as MarkupSections with tagName "li" instead of ListItems.

bantic:
This adds code to parse DOM with ul and ol correctly. Prior to this, markups with tagName "li" were incorrectly generated sometimes, and it was possible to create a corrupt mobiledoc by pasting in a list.

fixes #183

The parsing is fairly naive, and doesn't handle all types of nested lists. On those that it can't handle it just ignores the parsed sections rather than inserting invalid stuff into the mobiledoc, though.
There is a commented-out test for the type of nested lists that can appear in google docs HTML.

Pasting a list into content-kit can sometimes do nothing when the post to be inserted is a single list section. `PostEditor#insertPost` will incorrectly interpret that as a single markup section. This doesn't cause errors but it does prevent the list from being pasted. Pasting a list if it is part of a larger body (at least 1 other section) works correctly.

### DIFF
--- a/tests/unit/parsers/dom-test.js
+++ b/tests/unit/parsers/dom-test.js
@@ -323,4 +323,67 @@ test('unrecognized attributes are ignored', (assert) => {
   assert.ok(!markup.getAttribute('style'), 'style attribute not included');
 });
 
-// FIXME TODO ul, ol, li, img parsing
+test('singly-nested ul lis are parsed correctly', (assert) => {
+  let element= buildDOM(`
+    <ul><li>first element</li><li>second element</li></ul>
+  `);
+  const post = parser.parse(element);
+
+  assert.equal(post.sections.length, 1, '1 section');
+  let section = post.sections.objectAt(0);
+  assert.equal(section.tagName, 'ul');
+  assert.equal(section.items.length, 2, '2 items');
+  assert.equal(section.items.objectAt(0).text, 'first element');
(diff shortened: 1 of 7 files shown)


bantic:
Added #189 and #190 to address the extra work mentioned above.

tests/unit/parsers/dom-test.js
```diff
@@ -323,4 +323,67 @@ test('unrecognized attributes are ignored', (assert) => {
...
+  assert.equal(post.sections.length, 1, '1 section');
+  let section = post.sections.objectAt(0);
+  assert.equal(section.tagName, 'ul');
+  assert.equal(section.items.length, 2, '2 items');
+  assert.equal(section.items.objectAt(0).text, 'first element');
+  assert.equal(section.items.objectAt(1).text, 'nested element');
+});
+
+/*
+ * FIXME: Google docs nests uls like this
```
mixonic: punting on this for now?

bantic: Yeah. See #189. The immediate goal of this PR is to fix the problem where pasting a list (any list) could put the mobiledoc into a corrupt state (markups with "li" tagName outside of actual list sections).

mixonic:
@bantic drat, needs a rebase

PR #193, opened 2015-10-26
Add `cardParsers` to SectionParser

bantic:
This allows runtime-defined overrides for parsed content. This is mainly
useful for handling pasted content.

cc @mixonic

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -72,6 +72,7 @@ class Editor {
     this._elementListeners = [];
     this._views = [];
     this.isEditable = null;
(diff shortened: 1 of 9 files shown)


mixonic:
Not worth the bikeshed for now, but I think this would be neat as an API that gets every node. Not just cards. If you wanted to filter on cards you can use an `if` or `switch` in your hook to filter for that node. This would be similar to the AST preprocessing hooks on HTMLBars/Ember.

mixonic:
Looks great, let's just debate the above before documenting. Awesome!!

PR #200, opened 2015-11-02
Remove unused files, remove packages from package.json, docs

bantic:
Updates the docs for running the demo, removes several unused files and some dependencies and devDependencies.

cc @mixonic LMK if anything here looks like it shouldn't have been removed.


mixonic:
fantastic, thanks

PR #203, opened 2015-11-02
Enable pasting html and text externally

### LINKED ISSUE #180: Copy/Paste from Content-Kit to 3rd parties
Per this conversation, we should bring back HTML copy support to 3rd party paste targets. https://github.com/bustlelabs/content-kit-editor/pull/147/files#r42781191

bantic:
Sets the "text/html" and "text/plain" clipboard data on copy and cut
events.

fixes #180 

Since this uses textContent the plain-text copy data has no linebreaks, but that seems fine as a first approach.

### DIFF
--- a/src/js/utils/paste-utils.js
+++ b/src/js/utils/paste-utils.js
@@ -1,9 +1,8 @@
 /* global JSON */
 import MobiledocParser from '../parsers/mobiledoc';
 import HTMLParser from '../parsers/html';
-
-// FIXME fix this issue:
-const issueUrl = 'https://github.com/bustlelabs/content-kit-editor/issues/180';
+import { createDiv } from '../utils/element-utils';
+import { getSelectionContents } from '../utils/selection-utils';
 
 export function setClipboardCopyData(copyEvent, editor) {
   const { cursor, post } = editor;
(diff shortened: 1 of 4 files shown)


src/js/utils/paste-utils.js
```diff
@@ -11,11 +10,14 @@ export function setClipboardCopyData(copyEvent, editor) {
 
   const range = cursor.offsets;
   const mobiledoc = post.cloneRange(range);
+  const fragment = getSelectionContents();
+  const div = createDiv();
+  div.appendChild(fragment);
   const html =
     `<div data-mobiledoc='${JSON.stringify(mobiledoc)}'>` +
-      `<a href='${issueUrl}'>Pasting from Content-Kit not yet supported.</a>` +
+      div.innerHTML +
```
mixonic: I gave this some thought before- and I think it should perhaps copy a `mobiledoc-html-renderer` or `dom-renderer` version of the HTML.

* The `innerHTML` may have cards in an edit state, which would not be what we really want to copy.
* The DOM of the content-kit-editor is specialized, for example having empty `p` tags in a few cases where we need a place for the cursor. These tags are not really part of documents, however.

src/js/utils/paste-utils.js
```diff
@@ -11,11 +10,14 @@ export function setClipboardCopyData(copyEvent, editor) {
...
     `</div>`;
-  const plain = `Pasting from Content-Kit not yet supported. (${issueUrl})`;
+  const plain = div.textContent;
```
mixonic: And that implies this would be a new `mobiledoc-text-renderer`

src/js/utils/paste-utils.js
bantic: Using the `mobiledoc-html-renderer` or `mobiledoc-dom-renderer` here would definitely be more correct. I am reluctant to bring those in as deps for what seems like a fairly small use-case at the moment.

The use-case I'm targeting here is to make copy-pasting from a content-kit instance into something else non-surprising (as opposed to semantically accurate), so that copy/pasting out of content-kit and into google docs or medium displays something that looks right.

Because we hide a serialized mobiledoc in the clipboard data when copying, pasting from one content-kit editor instance to another is always a perfect reproduction (modulo any cards used that aren't present in the target's editor).

mixonic: IMO it is worth using the renderer for. Can we compromise by merging this in and I'll create a followup issue?

bantic: :+1: I'll create it

PR #210, opened 2015-11-03
Remove Post parser, its reparse methods are  now in DOM parser

bantic:
The "Post" parser is no longer used directly. It is now the `DOMParser`, and the reparsing code is part of it, now. It may be clearer to have a separate "Reparser" rather than a `reparse` method on the DOMParser.


mixonic:
:+1: to "dom-parser" "dom-reparser" or something

PR #213, opened 2015-11-04
add underline, subscript, superscript and strikethrough to allowed markup tags

vitosamson:


bantic:
@vitosamson Thanks, this is great. Could you squash the commits and drop `del` since it is a duplicate of `s`? (Eventually we'll also do the same for `b`/`i` since they are dupes of `strong`/`em`)

mixonic:
Superseded by https://github.com/bustlelabs/content-kit-editor/pull/217

vitosamson:
Why not allow the user to use either `del` or `s`? They have similar uses but are semantically different. http://html5doctor.com/ins-del-s/

Issue #222, opened 2015-11-10
Atoms


bantic:
As discussed with @rlivsey and @mixonic , we intend to implement "atoms", which are like markers except
  * are treated as though they have a length of 1, so cursors move across them, they are non-editable, and can be deleted as a unit
  * have metadata (like a card's payload) that requires a `text` property
  * Can use custom renderers (like cards)
  * Fallback to rendering only the text property

More notes are here: https://docs.google.com/document/d/1Wtuk5VG17lWwVSVuBH5tsdNgtbhDvWVt7qcEiVl3nqo/edit?usp=sharing

bantic:
closed by #311

PR #226, opened 2015-11-12
Atoms

rlivsey:
Work in progress thoughts on Atoms

Will implement #222 

* [x] render atom in renderers/editor-dom.js
* [x] parse v0.3 mobiledoc in parsers/mobiledoc.js
* [x] render v0.3 mobiledoc in renderers/mobiledoc.js
* [x] support atoms in text renderer - bustlelabs/mobiledoc-text-renderer#5
* [x] support atoms in DOM renderer - bustlelabs/mobiledoc-dom-renderer#20
* [x] support atoms in HTML renderer - bustlelabs/mobiledoc-html-renderer#15
* [x] atoms should set `contenteditable` to false
* [x] parse atoms in `DOMParser`
* [x] handle cursors (eg when no content before/after)
* [x] handle deletion

Maybe:

* [x] add default `unknownAtomHandler` implementation which just renders the value?

### DIFF
--- a/ATOMS.md
+++ b/ATOMS.md
@@ -1,0 +1,58 @@
+# Mobiledoc Atoms
+
+Atoms are effectively read-only inline cards.
+
+## Atom format
+
+An atom is a javascript object with 3 *required* properties:
+
+  * `name` [string] - The name of this atom in the mobiledoc
+  * `type` [string] - The output of this atom. Valid values are 'dom', 'html', and 'text'
+  * `render` [function] - Invoked by the renderer to render this atom
+
+## Atom rendering
+
+The `render` function on an atom is called by an instance of a renderer and passed an object with the following four properties:
+
+  * `env` [object] - A set of environment-specific properties
+  * `options` [object] - Rendering options that were passed to the renderer (as `cardOptions`) when it was instantiated
+  * `payload` [object] - The data payload for this atom from the mobiledoc
+  * `value` [string] - The textual representation to for this atom
+
+The return value of the `render` function will be inserted by the renderer into the rendered mobiledoc.
+The return value can be null if an atom does not have any output. If there is a return value it
+must be of the correct type (a DOM Node for the dom renderer, a string of html or text for the html or text renderers, respectively).
+
+#### `env`
+
+`env` always has the following properties:
+
+  * `name` [string] - the name of the card
+  * `onTeardown` [function] - The atom can pass a callback function: `onTeardown(callbackFn)`. The callback will be called when the rendered content is torn down.
+
+## Atom Examples
+
+Example dom atom that renders a mention:
+
+```js
+export default {
+  name: 'mention',
+  type: 'dom',
+  render({ env, options, value, payload}) {
+    return document.createTextNode(`@${value}`);
+  }
+};
+```
+
+Example dom atom that registers a teardown callback:
+```js
+let card = {
+ name: 'atom-with-teardown-callback',
+ type: 'dom',
+ render({env, options, value, payload}) {
+   env.onTeardown(() => {
+    console.log('tearing down atom named: ' + env.name);
+   });
+ }
+};
+```
--- a/ATOMS.md
+++ b/ATOMS.md
@@ -4,7 +4,7 @@ Atoms are effectively read-only inline cards.
 
 ## Atom format
 
-An atom is a javascript object with 3 *required* properties:
+An atom is a JavaScript object with 3 *required* properties:
 
   * `name` [string] - The name of this atom in the mobiledoc
   * `type` [string] - The output of this atom. Valid values are 'dom', 'html', and 'text'
--- a/MOBILEDOC.md
+++ b/MOBILEDOC.md
@@ -22,92 +22,152 @@ prescription for output display.
 Often Mobiledoc will be used with one or more of:
 
 * [Mobiledoc Kit](https://github.com/bustlelabs/mobiledoc-kit) to author an editor
-  * [Ember Mobiledoc Editor](https://github.com/bustlelabs/ember-mobiledoc-editor) is one such editor 
+  * [Ember Mobiledoc Editor](https://github.com/bustlelabs/ember-mobiledoc-editor) is one such editor
 * [Mobiledoc DOM Renderer](https://github.com/bustlelabs/mobiledoc-dom-renderer)
 * [Mobiledoc HTML Renderer](https://github.com/bustlelabs/mobiledoc-html-renderer)
 * [Mobiledoc Text Renderer](https://github.com/bustlelabs/mobiledoc-text-renderer)
 
 ## Specification
 
-Mobiledoc consists of a wrapping object and nested array of section data. Using
-arrays makes Mobiledocs each to render quickly.
+Mobiledoc consists of a wrapping object, type definitions for markup, atoms and cards,
+and an array of section data. Using arrays makes Mobiledocs each to render quickly.
 
 The wrapper signature:
 
 ```
 {
-  version: "0.1",                         ──── Versioning information
-  sections: [
-    [                                     ──── List of markup types.
-      markup,
-      markup
-    ],
-    [                                     ──── List of sections.
-      section,
-      section,
-      section
-    ]
+  version: "0.3",                         ──── Versioning information
+  markups: [                              ──── List of markup types
+    markup,
+    markup
+  ],
+  atoms: [                                ──── List of atom types
+    atom,
+    atom
+  ],
+  cards: [                                ──── List of card types
+    card,
+    card
+  ],
+  sections: [                             ──── List of sections.
+    section,
+    section,
+    section
   ]
 }
 ```
 
-**Markup signature**
+**Markup definition signature**
+
+Markups have a tagName, and optionally an array of `attributeName, attributeValue]` pairs
 
 ```
 {
-  version: "0.1",
-  sections: [
+  version: "0.3",
+  markups: [
     [tagName, optionalAttributes],        ──── Markup
     ['em'],                               ──── Example simple markup
     ['a', ['href', 'http://google.com']], ──── Example markup with attributes
-  ],[
-    // ...
-  ]]
+  ]
 }
 ```
 
-**Text Section**
+**Atom definition signature**
+
+Atoms have a name, text value and arbitrary payload.
 
 ```
 {
-  version: "0.1",
-  sections: [[
-    ["b"],                                ──── Markup at index 0
-    ["i"]                                 ──── Markup at index 1
-  ],[
-    [typeIdentifier, tagName, markers],   ──── typeIdentifier for text sections
-    [1, "h2", [                                is always 1.
-      [[], 0, "Simple h2 example"],
-    ]],
-    [1, "p", [
-      [openMarkupsIndexes, numberOfClosedMarkups, value],
-      [[], 0, "Example with no markup"],
-      [[0], 1, "Example wrapped in b tag"],
-      [[1], 0, "Example opening i tag"],
-      [[], 1, "Example closing i tag"],
-      [[1, 0], 1, "Example opening i tag and b tag, closing b tag"],
-      [[], 1, "Example closing b tag"]
-    ]]
-  ]]
+  version: "0.3",
+  atoms: [
+    [atomName, atomText, atomPayload],    ──── Atom
+    ['mention', '@bob', { id: 42 }]       ──── Example 'mention' atom
+  ]
 }
 ```
 
-The first item in the `sections` array is a list of markups. Markups have
-a tagName, and optionally an array of `attributeName, attributeValue]` pairs.
+**Card definition signature**
+
+Cards have a name and arbitrary payload.
+
+```
+{
+  version: "0.3",
+  cards: [
+    [cardName, cardPayload],            ──── Card
+    ['image', {                         ──── Example 'image' card
+      src: 'http://google.com/logo.png'
+    }]
+  ]
+}
+```
+
+**Markup Section**
+
+Markup sections, in addition to plain text, can include markups and atoms.
+
+```
+{
+  version: "0.3",
+  markups: [
+    ["b"],                                ──── Markup at index 0
+    ["i"]                                 ──── Markup at index 1
+  ],
+  atoms: [
+    ["mention", "@bob", { id: 42 }]       ──── mention Atom at index 0
+    ["mention", "@tom", { id: 12 }]       ──── mention Atom at index 1
+  ]
+  sections: [
+    [sectionTypeIdentifier, tagName, markers],   ──── sectionTypeIdentifier for markup sections
+    [1, "h2", [                                       is always 1.
+      [0, [], 0, "Simple h2 example"],
+    ]],
+    [1, "p", [
+      [textTypeIdentifier, openMarkupsIndexes, numberOfClosedMarkups, value],
+      [0, [], 0, "Example with no markup"],      ──── textTypeIdentifier for markup is always 0
+      [0, [0], 1, "Example wrapped in b tag"],
+      [0, [1], 0, "Example opening i tag"],
+      [0, [], 1, "Example closing i tag"],
+      [0, [1, 0], 1, "Example opening i tag and b tag, closing b tag"],
+      [0, [], 1, "Example closing b tag"],
+    ]],
+    [1, "p", [
+      [textTypeIdentifier, atomIndex, openMarkupsIndexes, numberOfClosedMarkups],
+      [1, 0, [], 0],                                    ──── mention atom at index 0 (@bob), textTypeIdentifier for atom is always 1
+      [1, 1, [0], 1]                                    ──── mention atom at index 1 (@tom) wrapped in b tag
+    ]],
+  ]
+}
+```
+
 The index in `openMarkupsIndex` specifies which markups should be opened at
 the start of the `value` text. As these tags are opened, then create a stack
 of opened markups. The `numberOfClosedMarkups` says how many markups should
 be closed at the end of a `value`.
 
+In addition to markups, markup sections may contain [ATOMS](ATOMS.md).
+Atoms have a `textTypeIdentifier` of 1 and contain a `atomTypeIndex`, text content
+and an `atomPayload` object which is arbitrary and passed through to the atom's
+implementation.
+
+Atoms also have `openMarkupsIndex` and `numberOfClosedMarkups` so that markup can flow
+across them.
+
+If an atom is present in Mobiledoc, but no atom implementation is registered, then the text
+value of the atom will be rendered as plain text as a fallback.
+
 **Card Section**
 
 ```
 {
-  version: "0.1",
-  sections: [[],[
-    [typeIdentifier, tagName, markers],   ──── typeIdentifier for card sections
-    [10, "card-name", cardPayload]             is always 10.
-  ]]
+  version: "0.3",
+  cards: [
+    ["card-name", { cardPayload }]
+  ],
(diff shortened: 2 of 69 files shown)


ATOMS.md
```diff
@@ -0,0 +1,63 @@
...
+Atoms are read-only so, unlike cards, do not have an `edit` hook.
+
+Each hook has a `setup` and `teardown` method. The arguments are:
+
+* `element` is a DOM element for that section. Nodes of that view for a atom
+  should be appended to the `element`.
+* `buffer` is an array passed to the `html` hook instead of a DOM element.
+  The content for the atom should be pushed on that array as a string.
+* `options` is the `cardOptions` argument passed to the editor or renderer.
+* `env` contains information about the running of this hook. It may contain
```
mixonic: If a card can remove itself, can it also save itself? It if can trigger its own destruction, it seems reasonable that it could change its payload as well.

On the other hand, part of me thinks it should not be able to remove or edit itself. The most conservative approach is that they can only be created and if they are removed it must be through normal editing operations like backspace and delete. Thoughts?

ATOMS.md
```diff
@@ -0,0 +1,63 @@
...
+    hook and removes the atom from DOM and from the post abstract.
+    the instance to edit mode.
+* `text` is the textual value used to display to the user.
+* `payload` is the payload for this atom instance. It was either loaded from
+  a Mobiledoc or set when the atom was inserted.
+
+Additionally, *renderers may offer the ability to configure a non-standard
+hook name at runtime*. An example would be having the DOM renderer called with
+an option specifying the hook name `mobile-placeholder`. This allows for
+variants of an atom in different situations.
```
mixonic: This describes something similar we already have for cards, right? We should better describe and document that system in general IMO. We need to unify the idea of `html`, `display`, `edit`, `text` etc. Can punt for now though.

MOBILEDOC.md
```diff
@@ -96,6 +96,29 @@ the start of the `value` text. As these tags are opened, then create a stack
...
+{
+  version: "0.3",
+  sections: [[
+    ["b"],                                ──── Bold Markup at index 0
+    ["i"],                                ──── Italic Markup at index 1
+    ["atom", "mention"]                   ──── Mention Atom at index 2
+  ],[
+    [openMarkupsIndexes, numberOfClosedMarkups, value, payload],
+    [[2], 0, "Bob", { id: 42 }],          ──── Mention Atom with payload
+    [[1], 0, "Example opening i tag"],
```
mixonic: we chatted about this in chat. IMO atoms should not be implemented as markups.

ATOMS.md
rlivsey: I took out all the other ones from cards because save/cancel/edit/section didn't seem to be needed for atoms at first glance. 

I can picture a use-case for `remove` such as showing an `x` on hover and letting you click it to remove the atom, but certainly happy to leave that out initially!

If we do have remove, I can think of some use-cases where an atom could be expose an interface which alters it and therefor would want to update its payload. Not something I personally need right now hence why it didn't spring to mind initially.

ATOMS.md
rlivsey: Yep, that's all copied & pasted from the card docs! We can probably merge the atoms & card docs together as there are certainly commonalities vs having them in two files.

MOBILEDOC.md
rlivsey: From chat:

> @rlivsey hm, I imagined that they were a peer of markers
> and not markup at all
> it might require a type for markers

```
[0 /* means marker */, [2], 0, 'bob'],
[1 /* means atom */, 'mention', 'martin', { url: 'google.com }]
```

So split out the atom definitions into their own array too?

Eg something like:

```
{
  version: "0.3",
  sections: [[
    ["b"],         ──── Bold Markup at index 0
    ["i"],         ──── Italic Markup at index 1
  ],[
    ["mention"]    ──── Mention Atom at index 0
    ["thingy"]     ──── Thingy Atom at index 1
  ],[
    [0, [2], 0, 'bob'],                             ──── Italic Markup
    [1, 'mention', 'martin', { url: 'google.com' }] ──── Mention Atom
  ]]
}
```

As this is backwards incompatible anyway, would it be worth moving the definition to their own parts of the mobiledoc so the sections array is purely for the content?

Eg something like:

```
{
  version: "0.3",
  markups: [
    ["b"],         ──── Bold Markup at index 0
    ["i"],         ──── Italic Markup at index 1
  ],
  atoms: [
    ["mention"]    ──── Mention Atom at index 0
    ["thingy"]     ──── Thingy Atom at index 1
  ],
  sections: [
    [0, [2], 0, 'bob'],                             ──── Italic Markup
    [1, 'mention', 'martin', { url: 'google.com' }] ──── Mention Atom
  ]
}
```

ATOMS.md
mixonic: `remove` with an `x` implies that the atom is (partially) in charge of its editing interface. I'd prefer to lean on the conservative path for now and ensure they have *zero* control over their editing interface.

We can revisit as we get more complex use-cases, if that seems reasonable to you as well.

rlivsey: :+1:

MOBILEDOC.md
bantic: I've wanted to separate sections from markups for a while now, so I am +1 to the above.

The `sections` you have above are not real sections, though. I think you meant to wrap in a section, so it should be:
```
sections: [1, "P", [                                  ——— a "p" markup section containing two markers:
  [0, [2], 0, 'bob'],                                    ─ ─ ─ marker type 0 (normal marker), italic
  [1, 0, 'martin', { url: 'google.com' } ]          ─ ─ ─ marker type 1 (atom), atom index 0 ('mention'), text, payload
]]
```

rlivsey: Cool, yep that was cut & paste fail :grinning:

ATOMS.md
```diff
@@ -0,0 +1,63 @@
...
+
+Atoms are effectively read-only inline cards.
+
+An atom is an object with a name and at least one of several hooks defined. For example:
+
+```js
+var demoAtom = {
+  name: 'mention',
+  display: {
+    setup(element, options, env, text, payload) {
```
mixonic: I know this is annoying, but let's change this to follow the new card schema as suggested in https://github.com/bustlelabs/mobiledoc-kit/issues/228#issuecomment-156578655

```js
// package: mda-profile-dom
export default {
  name: 'profile',
  type: 'dom',
  render({element, options, env, text, payload}) {
  }
};
```

ATOMS.md
```diff
@@ -0,0 +1,63 @@
...
+Like [CARDS](CARDS.md) there are several hooks an atom should define.
+
+```js
+var exampleAtom = {
+  name: 'example',
+  display: {
+    setup(element, options, env, text, payload) {},
+    teardown(setupReturnValue) {}
+  },
+  html: {
```
mixonic: Right, and according to the above these would be two atoms instead of one with two modes:

```js
// package: mda-example
// file: dom.js
export default {
  name: 'example',
  type: 'dom',
  render({element, options, env, text, payload}) {
  }
};
```

```js
// package: mda-example
// file: html.js
export default {
  name: 'example',
  type: 'html',
  render({buffer, options, env, text, payload}) {
  }
};
```

MOBILEDOC.md
```diff
@@ -22,92 +22,149 @@ prescription for output display.
...
 ```
 
-The first item in the `sections` array is a list of markups. Markups have
-a tagName, and optionally an array of `attributeName, attributeValue]` pairs.
 The index in `openMarkupsIndex` specifies which markups should be opened at
 the start of the `value` text. As these tags are opened, then create a stack
 of opened markups. The `numberOfClosedMarkups` says how many markups should
 be closed at the end of a `value`.
 
+In addition to markups, text sections may contain [ATOMS](ATOMS.md).
```
mixonic: We can keep calling them "markup sections" for now.

README.md
```diff
@@ -130,10 +130,21 @@ It is important that you make changes to posts, sections, and markers through
...
 and better understand changes being made to the post.
 
+```js
+editor.run(postEditor => {
+  const mention = postEditor.builder.createAtom("mention", "John Doe", { id: 42 });
+  // insert at current cursor position:
+  // or should the user have to grab the current position from the editor first?
+  postEditor.insert(mention);
+  // or specify a different position:
+  postEditor.insert(mention, position);
```
mixonic: I worry about `insert` here, especially since we already have a pretty heavily used `insertBefore` that uses a different signature and his a different meaning. As we get closer to implementing this part of the API I suggest pondering alternatives.

mixonic:
Looks good to go :-) @rlivsey I presume we will just leave it open while you spike on implementation this week.

We should attempt to preserve the 0.2 compatibility as we make changes. It will be a bit more painful in the short term since the versioning of parsers/renderer will need to be implemented, but will allow us to merge patches incrementally.

README.md
rlivsey: Agreed. 

A `replace` API could be of use for atoms too, as a common use-case would be to insert an atom over existing content eg after typing `@bob` in `lorem ipsum @bob` you'd want to replace the mention text which had just been entered with the new atom. 

I could potentially see that being more common than inserting an atom, certainly for the @mention use-case anyway. Pretty much just syntactic sugar on top of delete & insert but would remove boilerplate.

src/js/models/atom-node.js
```diff
@@ -0,0 +1,42 @@
...
+
+  get env() {
+    return {
+      name: this.atom.name,
+      model: this.model
+    };
+  }
+
+  remove() {
+    // this.editor.run(postEditor => postEditor.removeSection(this.model));
```
mixonic: should likely teardown here

src/js/models/atom-node.js
```diff
@@ -0,0 +1,42 @@
...
+      env: this.env,
+      text: this.model.text,
+      payload: this.model.payload
+    });
+  }
+
+  get env() {
+    return {
+      name: this.atom.name,
+      model: this.model
```
mixonic: `model` -> `atom`

mixonic: Yeah lets just drop this

src/js/models/atom-node.js
mixonic: This can just be removed.

src/js/renderers/editor-dom.js
```diff
@@ -91,6 +93,21 @@ function renderCard() {
   return { wrapper, cardElement };
 }
 
+function renderAtom(element, previousRenderNode) {
+  let atomElement = document.createElement('span');
+  addClassName(atomElement, 'ck-atom');
```
mixonic: do we need a class name here?

At the least let's make it `-mobiledoc-kit__atom`

src/js/models/atom-node.js
```diff
@@ -0,0 +1,42 @@
...
+    this.model = model;
+    this.atomOptions = atomOptions;
+    this.element = element;
+
+    this.setupResult = null;
+  }
+
+  render() {
+    this.teardown();
+
```
mixonic: Here we've discussing passing a document fragment instead. Pass it, then after `render` attach it to the `element`

src/js/models/atom-node.js
```diff
@@ -0,0 +1,42 @@
...
+      model: this.model
+    };
+  }
+
+  remove() {
+    // this.editor.run(postEditor => postEditor.removeSection(this.model));
+  }
+
+  teardown() {
+    // TODO - test if setupResult is a function before calling?
```
mixonic: And in teardown, we should clear all children of the `element`

tests/unit/renderers/editor-dom-test.js
```diff
@@ -172,6 +172,27 @@ test('renders a post with image', (assert) => {
...
 });
 
+test('renders a post with atom', (assert) => {
+  let post = builder.createPost();
+  let section = builder.createMarkupSection('P');
+  post.sections.append(section);
+
+  section.markers.append(
+    builder.createAtom('mention', '@bob', {})
+  );
```
bantic: This can be done a bit more simply with the test helper for postAbstract:
```
let post = Helpers.postAbstract.build(builder => {
  let { markupSection, post, atom } = builder;
  return post([markupSection('p', [atom('mention', '@bob', {})])]);
});
```

tests/unit/models/atom-test.js
```diff
@@ -0,0 +1,24 @@
...
+
+test('can create an atom with text and payload', (assert) => {
+  const payload = {};
+  const text = 'atom-text';
+  const name = 'atom-name';
+  const atom = builder.createAtom(name, text, payload);
+  assert.ok(!!atom, 'creates atom');
+  assert.ok(atom.name === name, 'has name');
+  assert.ok(atom.text === text, 'has text');
+  assert.ok(atom.payload === payload, 'has payload');
```
bantic: We should probably also add a `length` property to an atom that returns `1`. Worth testing that here, and also worth testing that a markup section calculates its length correctly with a mix of text markers and atoms.

rlivsey: That would make sense, as soon as I mix markup with atoms that's the first failure I get as current code expects a length property!

bantic:
This looks good — probably the first set of challenges will be breakage caused by methods in [`models/_markerable`](https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/models/_markerable.js) expecting to always encounter markers. 
The tests are a little thin around this area, but the [unit tests for markupSection](https://github.com/bustlelabs/mobiledoc-kit/blob/master/tests/unit/models/markup-section-test.js) would be the right spot to add some more tests that those methods work properly when there is a mix of atoms and markers.

rlivsey:
Getting close…

ATOMS.md
```diff
@@ -0,0 +1,58 @@
+# Mobiledoc Atoms
+
+Atoms are effectively read-only inline cards.
+
+## Atom format
+
+An atom is a javascript object with 3 *required* properties:
```
mixonic: "JavaScript" nit pick :-p

README.md
```diff
@@ -55,7 +59,7 @@ editor.render(element);
 * `spellcheck` - [boolean] whether to enable spellcheck. Defaults to true.
 * `autofocus` - [boolean] When true, focuses on the editor when it is rendered.
 * `cards` - [array] The list of cards that the editor may render
-* `cardOptions` - [object] Options passed to 
+* `cardOptions` - [object] Options passed to
```
mixonic: "Options passed to cards"

src/js/models/atom-node.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  _validateAndAppendRenderResult(rendered) {
+    if (!rendered) {
+      return;
+    }
+
+    let { atom: { name } } = this;
+    assert(
+      `Atom "${name}" must render dom (render value was: "${rendered}")`,
```
mixonic: "must return a DOM node" ?

rlivsey:
Atoms now have ZWNJ's around them, so when in a section on their own the cursor appears correctly.

Next up is figuring out that the cursor is directly before/after the atom so we can jump across when hitting left/right. For cards this is easy enough as the whole section is a card and so `position.section.isCardSection` is all that's needed.

Detecting if you're next to an atom is a little trickier by the looks of it.

The atom's DOM looks like:

```html
<span>
  ZWNJ
  <span class="-mobiledoc-kit__atom">...</span>
  ZWNJ
</span>
```

If the cursor is right after the atom, or in the `ZWNJ`, then it looks like `this.cursor.selection.anchorNode` is the `ZWNJ` and we can do something like this to see if we're in the atom:

```javascript
let renderNode = this._renderTree.getElementRenderNode(anchorNode.parentNode);
if (renderNode && renderNode.postNode.isAtom) {
   // ...
}
```

If the cursor is directly before the atom however, it's not in the `ZWNJ`, at this point we could do something like:

```javascript
let _comparePosition = comparePosition(this.cursor.selection);
if (_comparePosition.tailNode.length === _comparenPosition.tailOffset) { // at the end of the node
  let nextNode = _comparePosition.tailNode.nextElementSibling;
  let renderNode = nextNode && this._renderTree.getElementRenderNode(nextNode);
  if (renderNode && renderNode.postNode.isAtom) {
     // ...
  }
}
```

Any thoughts on a nicer way of handling this?

mixonic:
One test away from passing, though I think deletion is still pending :-) :-) :-)

rlivsey:
@mixonic awesome!

src/js/models/atom-node.js
mixonic: punting.

mixonic:
Rebased this to be a bit more clean commit-wise, and it seems ready for master to me.

bantic:
Nice! Very cool to see and use this in action. Monumental amount of work involved here...
I noticed a few things when smoke testing this that should probably be fixed up before merging:
  * hitting enter within the section caused an error (related to attempting to call `marker.clone` — perhaps just needs to implement `atom.clone`)
  * selecting only the atom and clicking the bold/italic buttons caused odd things to happen — looks like something in the render tree not getting marked dirty, because parts of the rendered text disappeared later when typing new text

Really cool. I will look more closely at the diff and give feedback tomorrow.

mixonic:
@bantic your just-mentioned issues should now be addressed. Thanks :clap: :-)

bantic:
@mixonic Awesome! I was able to find a few new things here, and would be happy to jump in and help fix them:
  * seems to be possible to type text in the end-position zwnj cursor-placeholder that shows up in dom but is not reflected in the mobiledoc. I navigated to it using the keyboard arrow keys
  * selecting only the atom (place cursor on left side, use shift + right-arrow to select across) and modifying markup (click "Bold" button or press cmd-B) seems to put the editor into a funny state — a linebreak is added

bantic:
a sample mobiledoc that appears to render incorrectly (corresponds to clicking "bold" when only an atom is selected): https://gist.github.com/bantic/cba7c8a6ed50e8407b76

bantic:
Did some work to handle typing at the end of the atom in https://github.com/bustlelabs/mobiledoc-kit/tree/atoms-with-end-cursor-positioning.
More notes coming.

bantic:
I am finding one outstanding bug that I'm working on fixing:
  * [ ] select only atom, make it bold (=> rendering errors/inconsistencies, e.g. type character at left side of bolded atom)

And have fixed the following bugs that now need test coverage:
  * [ ] type text between two atoms
  * [ ] before 1st atom: type text (ensure cursor is correct place)
  * [ ] after last atom: type text (ensure cursor is in correct place)
  * [ ] hitting left arrow in at start of 1st atom in section (should go to end of prev section)
  * [ ] hitting right arrow at end of last atom in section
  * [ ] hitting right arrow at left of last atom -> cursor does not move to right side of atom (cursor._findNodeForPosition was finding the wrong node for the position)
  * [ ] ensure the rendered output after `editor#reparse` when there is text content in the atom zwnj is correct output (must mark the section as dirty so that the newly-inserted marker gets rendered)
  * [ ] ctr-A does not move to start of atom when atom is at beginning of line
  * [ ] ctr-E does not move to end of line when cursor is left (anywhere left of) an atom
  * [ ] “Hello Bobbob” => write test to ensure that neighboring atoms are not coalesced
  * [ ] FIXED, needs tests: deleting text from right side of the atom => hitting delete when 1 character after right side deletes and places cursor on left side of atom (this is fixed, but I forget how)
  * [ ] select text that ends in the middle of two “Hello Bob” atoms and delete => atoms become “Hello B”. (section.markersFor does not work properly with atoms (it cuts their value in pieces))

mixonic:
Closing in favor of https://github.com/bustlelabs/mobiledoc-kit/pull/311

Issue #228, opened 2015-11-13
Card/Renderer unification RFC


bantic:
### Existing methods and inconsistencies

We now have published, quasi-official renderers for [dom](https://github.com/bustlelabs/mobiledoc-dom-renderer), [html](https://github.com/bustlelabs/mobiledoc-html-renderer) and [text](https://github.com/bustlelabs/mobiledoc-text-renderer)

When rendering cards, these renders have different expectations for named hooks that contain `setup` and `teardown` methods:

renderer: | dom | html | text
-----|------|--------|------
card property name: | `"display"` and `"edit"` | `"html"` | `"text"`

Since the DOM renderer is used both by the editor (to render cards in either "display" or "edit" mode on the editor surface) and for display-only environment (i.e., rendering cards that are displayed to a user reading an article on a site), the `display#setup` hook has special semantics for inferring its rendering environment based on the presence of an `edit` property in the `env` argument to `setup()`:

```javascript
let card = {
  name: 'some-card',
  display: {
    setup(element, options, env, payload) {
      if (env.edit) { // <-- the `edit` env property is only provided when rendering
                           // the card for the editor
        // do stuff that only happens in the editor, like showing an
        // "edit" or "remove" button
      } else {
        // do stuff that only happens when displayed to an end-user on a web page
      }
    }
  }
};
```

Another inconsistency is in the way the `setup` method works for the different renderers:

renderer: | dom | html | text
---------|------|-------|------
setup method 1st arg: | `element` | `buffer` | n/a
how to "render" : | create dom nodes, append to `element` | call `buffer.push(string)` repeatedly to add HTML strings to the output | return a string (the 1st arg to `setup` is unused) 

### Suggestions for unification

 * **Rename hooks to match renderers:** rename the `display` hook to `dom` so that it matches the name of its rendering context
 * **Clarify the difference between dom-rendering for the editor and for display-only situations:** Either create new hooks for the editor (unsure what is best, perhaps `editor-display` and `editor-edit`) or add a more expressive property to the env like `context` or `isInEditor` that a card can use to determine whether to show editor-specific UI (instead of inferring from the presence of `env.edit`)
 * **Unify the 1st argument to `setup`:** I propose a `buffer` object that only has a single method `push`. In dom rendering scenarios, for each node you wanted to append to the card you'd call `buffer.push(node)`. In html rendering scenarios you'd `push` a string of HTML. In text-rendering scenarios you'd `push` a string of text.

[...]

PR #234, opened 2015-11-17
Card Unification RFC v2

bantic:
This is a continuation/distillation of the helpful feedback from #228.

**do not merge**

### [rendered](https://github.com/bustlelabs/mobiledoc-kit/blob/card-unification-rfc/card-unification.md)

A few notes:
  * ember-based consumers that use `ember-mobiledoc-editor` or `ember-mobiledoc-dom-renderer` will not need to change much — the changes involved here are primarily for 1st-level consumers
  * The changes described above de facto mean that cards will be separated by their render context — the same card will not be used for multiple rendering contexts

### DIFF
--- a/card-unification.md
+++ b/card-unification.md
@@ -1,0 +1,90 @@
+# Card Unification
+
+  * [ ] Change card shape to object with `type`, `name`, `render` and optional `edit` properties
+  * [ ] Card's `type` is validated by renderer (dom renderer cannot render 'text', e.g.)
+  * [ ] Change arguments passed by editor to card's `render` (or `edit`) method
+    * [ ] single argument object with `name`, `buffer`, `isInEditor`, `options`, `editorEnv` and `payload` properties
+  * [ ] Return value of card's `render` (or `edit`) method is appended/concatenated by renderer
+  * [ ] card can register teardown callback by calling `buffer.onTeardown(teardownFn)`
+  * [ ] Change editor-dom renderer to clear child elements from card element on teardown
+  * [ ] Renderer `constructor` signature changes to accept cards (array), cardOptions (object) arguments
+  * [ ] Renderer `render` instance method accepts `mobiledoc` argument, returns `rendered` object with 2 props:
+    * [ ] `result` property (which is a dom node or string, depending)
+    * [ ] `teardown` method with no args tears down rendered mobiledoc (removing dom when applicable, calling registered card teardown callbacks when applicable)
+
+## Projects that must change
+
+ * [ ] [mobiledoc-html-renderer](https://github.com/bustlelabs/mobiledoc-html-renderer)
+ * [ ] [mobiledoc-dom-renderer](https://github.com/bustlelabs/mobiledoc-dom-renderer)
+ * [ ] [mobiledoc-text-renderer](https://github.com/bustlelabs/mobiledoc-text-renderer)
+ * [ ] editor-dom renderer within [mobiledoc-kit](https://github.com/bustlelabs/mobiledoc-kit)
+ * [ ] [ember-mobiledoc-editor](https://github.com/bustlelabs/ember-mobiledoc-editor)
+ * [ ] demo app within mobiledoc-kit
+
+## Card shape
+
+Every card is an object that *must* define the following properties:
+ * `render`: A function called by the renderer to display the card
+ * `type`: A string. The render context (currently the only known render contexts are 'dom', 'html' and 'text')
+ * `name`: A string. This is the same string that is in DOM
+
+Additionally a card *may* define the following property:
+
+ * `edit`: A function that the editor-dom renderer calls to render the card in edit state. Non-editor renderers will likely have no reason to have an edit property
+
+## Card `render` method
+
+The `render` method (and the `edit` method if applicable) receives a single argument that is an object with the following properties:
+
+  * `name`
+  * `buffer` An object supplied by the renderer that provides an `onTeardown` method to register a teardown callback
+  * `options` The card options passed to the renderer
+  * `isInEditor` True when being rendered in an editor, false otherwise
+  * `editorEnv`
+    * editorEnv will be an object populated, when being rendered by an editor, with the following properties:
+    * `save` Method with the signature `(newPayload, transition=false)`. Replaces payload with `newPayload` and transitions to display mode if `transition` is true
+    * `cancel` Method with no args. Exit edit mode without saving payload
+    * `edit` Method with no args. Change from display mode to edit mode
+    * `remove` Method with no args. Remove this card from the post abstract
+    * `section` The post abstract cardSection for this card (can be used with the `postEditor` to manipulate this card in other ways)
+  * `payload` The payload in the mobiledoc for this card
+
+### Rendering a card
+
+The **return value** of the `render` method is the rendered result of the card. This should either be a dom node (for cards with type 'dom') or a string (for cards with type 'html' or 'text'). Renderers should validate that the return value is the correct type.
+
+### Tearing down a card
+
+The 80% use case is likely that most cards will not need to clean up after themselves. The editor-dom renderer will clear child nodes, which should account for most use cases.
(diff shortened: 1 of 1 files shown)


card-unification.md
```diff
@@ -0,0 +1,90 @@
+# Card Unification
+
+  * [ ] Change card shape to object with `type`, `name`, `render` and optional `edit` properties
+  * [ ] Card's `type` is validated by renderer (dom renderer cannot render 'text', e.g.)
+  * [ ] Change arguments passed by editor to card's `render` (or `edit`) method
+    * [ ] single argument object with `name`, `buffer`, `isInEditor`, `options`, `editorEnv` and `payload` properties
```
mixonic: `isInEditor` -> `isEditing`? "in editor" seems funky.

I'm kind of inclined to put it on `env`, but I'll let is go :-p Just don't want to make this a junk drawer.

card-unification.md
```diff
@@ -0,0 +1,90 @@
...
+  * [ ] Return value of card's `render` (or `edit`) method is appended/concatenated by renderer
```
mixonic: Still a pretty strong disagree with this. iOS rendering, for example, would almost certainly not feature something that was to be returned. Adopting a pattern that feels nice today because it works well for returning strings seems short-sighted.

card-unification.md
```diff
@@ -0,0 +1,90 @@
...
+  * [ ] card can register teardown callback by calling `buffer.onTeardown(teardownFn)`
```
mixonic: Whoops, kind of confused. `buffer` is not a buffer at all, but is a spot to register teardown?

card-unification.md
```diff
@@ -0,0 +1,90 @@
...
+  * [ ] Change editor-dom renderer to clear child elements from card element on teardown
+  * [ ] Renderer `constructor` signature changes to accept cards (array), cardOptions (object) arguments
```
mixonic: cards, atoms, cardOptions

What happens to `editor` and `unknownCardHandler`? Is this the `renderer` here: https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/renderers/editor-dom.js#L360?

card-unification.md
```diff
@@ -0,0 +1,90 @@
...
+ * [ ] editor-dom renderer within [mobiledoc-kit](https://github.com/bustlelabs/mobiledoc-kit)
+ * [ ] [ember-mobiledoc-editor](https://github.com/bustlelabs/ember-mobiledoc-editor)
+ * [ ] demo app within mobiledoc-kit
+
+## Card shape
+
+Every card is an object that *must* define the following properties:
+ * `render`: A function called by the renderer to display the card
+ * `type`: A string. The render context (currently the only known render contexts are 'dom', 'html' and 'text')
+ * `name`: A string. This is the same string that is in DOM
```
mixonic: "in DOM" -> "in the mobiledoc"?

card-unification.md
```diff
@@ -0,0 +1,90 @@
...
+  * `buffer` An object supplied by the renderer that provides an `onTeardown` method to register a teardown callback
+  * `options` The card options passed to the renderer
+  * `isInEditor` True when being rendered in an editor, false otherwise
+  * `editorEnv`
+    * editorEnv will be an object populated, when being rendered by an editor, with the following properties:
+    * `save` Method with the signature `(newPayload, transition=false)`. Replaces payload with `newPayload` and transitions to display mode if `transition` is true
+    * `cancel` Method with no args. Exit edit mode without saving payload
+    * `edit` Method with no args. Change from display mode to edit mode
+    * `remove` Method with no args. Remove this card from the post abstract
+    * `section` The post abstract cardSection for this card (can be used with the `postEditor` to manipulate this card in other ways)
```
mixonic: I think we should take this change to change this to `postModel` or `postNode`. On `atoms` @rlivsey ran into `section` and had to pick a new name, I suggest they should be the same and this is a good time to change them.

card-unification.md
bantic: Yeah, the name `buffer` is misleading. This should instead just be a named property `onTeardown` that you can call. I'll change.

card-unification.md
bantic: I considered `isEditing` but to me it implies that the card is in "edit mode" (not "display mode"). `isInEditor` is intended to capture the fact that the card is being rendered for an editor (whether it is in edit or display mode at the moment). Same thing for `editorEnv` — that is an object of properties that could be useful when a card is being displayed/edited in an editor. `editorEnv.save` seems like it would be useful any time for a card in an editor (`save(newPayload, false)` is the way to update a card's payload without transitioning a card). `editorEnv.edit` is a no-op when the card is already in edit mode. It's redundant but doesn't seem confusing to me — I wouldn't expect someone to think that calling "edit" again while in edit mode would do something.

mixonic: Ah ok, I think I had the entire meaning of `isInEditor` wrong. Seems ok, I'm not sure what a better name would be :+1: 

`editorEnv` has some limited items like `name` for cards not rendered by the editor.

card-unification.md
bantic: I'm not sure we will be able to anticipate every alternative render context with this pattern, but for environments like iOS that aren't using JavaScript anyway I think it would be fine to break with this pattern and use an alternative SDK for those. I am reluctant to change the semantics of `render` to accommodate at-this-point-hypothetical alternative rendering contexts.

The thing I would like to strive for here is consistency among all JavaScript renderers — if using the return val is not the right fit then the best second option I can think of is having a buffer object that has a consistent API among all renderers (`buffer.push(stringOrDOMNode)`, or `buffer.append(stringOrDOMNode)`).

card-unification.md
bantic: `editor` is only necessary for the editor-dom renderer (there is no editor when using the other renderers).

`unknownCardHandler` needs to be accounted for. Currently only the editor-dom renderer has any concept of it. We should codify it for all renderers. I'll update this accordingly.

Given that a renderer may (optional) require an editor, and will need `cards`, `cardAtoms`, `cardOptions` and `unknownCardHandler` I am inclined to change the signature of the renderer constructor to accept an `options` argument instead of this increasing list of positional arguments.

bantic: I'll update this RFC accordingly.

card-unification.md
bantic: yep, typo. thanks.

card-unification.md
bantic: `postModel` sounds fine. What does an atom receive? Is it also be a section? (a markerable section like markupSection or listItem?)

mixonic: It should receive an `Atom` instance, similar to a `Marker`. Not a section at all, hence the name change.

bantic:
updated based on feedback

card-unification.md
```diff
@@ -0,0 +1,100 @@
+# Card Unification
+
+  * [ ] Change card shape to object with `type`, `name`, `render` and optional `edit` properties
+  * [ ] Card's `type` is validated by renderer (dom renderer cannot render 'text', e.g.)
+  * [ ] Change arguments passed by editor to card's `render` (or `edit`) method
+    * [ ] single argument object with `name`, `onTeardown`, `isInEditor`, `options`, `editorEnv` and `payload` properties
+  * [ ] Return value of card's `render` (or `edit`) method is appended/concatenated by renderer
```
gdub22: Looks like the other comments got removed here upon update.  My response:

This doesn't make a huge difference either way, but the idea is to simplify things . Leave the rendering to the actual render, instead of doing it in the card.

 - Saves writing `appendChild`, concatenation, `addSubview` code in every card, and just let's the renderer do the right thing.
 - `appendChild` abstractions can be kept in the renderer instead of re-implementing that in cards
 - The current pattern forces using wrapper dom element for cards.
 - Passing in a buffer element opens easier possibilities of users leaking it

I don't understand the iOS argument.  You return a UIView and the renderer does `addSubview`, instead of writing that line in every card.

bantic: :+1: Thanks for the feedback @gdub22. This makes sense, and I'm going to move forward with updating the renderers to work this way (using the return val as the rendered content and handling appending/concatting in the renderer itself)

bantic:
implemented in #235

PR #237, opened 2015-11-18
Retain edit mode on cloned cards

mixonic:
In Typeset, @ZeeJab was moving around new cards with the arrow buttons. This caused a card in edit mode to be cloned and moved to the new location. Because initial mode was not set, the card would flop to display and out of edit without a valid payload.

It seems reasonable that a new card would retain the current mode of render as you move it around. This adds logic to `clone` that set the initial mode of the new card correctly.

### DIFF
--- a/src/js/models/card.js
+++ b/src/js/models/card.js
@@ -24,8 +24,16 @@ export default class Card extends Section {
   }
 
   clone() {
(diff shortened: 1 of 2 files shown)


ZeeJab:
LOOKS GOOD TO Z! THANK YOU!! :game_die:

Issue #235, opened 2015-11-18
implement card rendering refactor


bantic:
This is a meta-issue that tracks implementing the changes explicated in #234. The notes below are copied from that PR.

## Next steps:

 * [x] release new version of mobiledoc-kit (0.6.2-beta1)
 * [x] update ember-mobiledoc-editor to use new mobiledoc-kit version
 * [x] release new version of ember-mobiledoc-editor (0.2.2-beta1)
 * [x] update demo to use released version of ember-mobiledoc-editor
 * [x] update documentation of card and renderer usage #245 
 * [x] update other projects using old card shapes (tracked separately)

## Projects finished being refactored

 * [x] editor-dom renderer within [mobiledoc-kit](https://github.com/bustlelabs/mobiledoc-kit) #236
 * [x] [mobiledoc-dom-renderer](https://github.com/bustlelabs/mobiledoc-dom-renderer/issues/17)
 * [x] [ember-mobiledoc-dom-renderer](https://github.com/bustlelabs/ember-mobiledoc-dom-renderer/issues/5)
 * [x] [ember-mobiledoc-editor](https://github.com/bustlelabs/ember-mobiledoc-editor/issues/33)
 * [x] [mobiledoc-html-renderer](https://github.com/bustlelabs/mobiledoc-html-renderer/issues/12)
 * [x] [mobiledoc-text-renderer](https://github.com/bustlelabs/mobiledoc-text-renderer)
 * [x] demo app within mobiledoc-kit #239
 * [x] update mobiledoc-html-renderer dep for ember-mobiledoc-html-renderer
 * [x] update mobiledoc-text-renderer dep for ember-mobiledoc-text-renderer

# Card Unification

  * Change card shape to object with `type`, `name`, `render` and optional `edit` properties
  * Card's `type` is validated by renderer (dom renderer cannot render 'text', e.g.)
  * Change arguments passed by editor to card's `render` (or `edit`) method
    * single argument object with `env`, `options` and `payload` properties
  * Return value of card's `render` (or `edit`) method is appended/concatenated by renderer
  * card can register teardown callback by calling `onTeardown(teardownFn)`
  * Change editor-dom renderer to clear child elements from card element on teardown
  * Renderer `constructor` signature changes to accept options object with: cards, atoms, cardOptions, unknownCardHandler, editor (for editor-dom)
  * Renderer `render` instance method accepts `mobiledoc` argument, returns `rendered` object with 2 props:
[...]

PR #238, opened 2015-11-18
Implement card refactor for editor-dom renderer

### LINKED ISSUE #236: Update editor-dom renderer for card rendering
see #235 

  * [x] Change card shape to object with `type`, `name`, `render` and optional `edit` properties
  * [x] Card's `type` is validated by renderer (dom renderer cannot render 'text', e.g.)
  * [x] Change arguments passed by editor to card's `render` (or `edit`) method
    * [x] single argument object with `env`, `options` and `payload` properties
  * [x] Return value of card's `render` (or `edit`) method is appended/concatenated by renderer
  * [x] card can register teardown callback by calling `onTeardown(teardownFn)`
  * [x] Change editor-dom renderer to clear child elements from card element on teardown
  * [ ] ~~Renderer `constructor` signature changes to accept options object with: cards, atoms, cardOptions, unknownCardHandler, editor (for editor-dom)~~ (not applicable since the editor-dom renderer is managed internally by mobiledoc-kit)
  * [ ] ~~Renderer `render` instance method accepts `mobiledoc` argument, returns `rendered` object with 2 props:~~
    * [ ] ~~`result` property (which is a dom node or string, depending)~~
    * [ ] ~~`teardown` method with no args tears down rendered mobiledoc (removing dom when applicable, calling registered card teardown callbacks when applicable)~~
  * [x] `unknownCardHandler` method is called whenever a renderer encounters an unknown card
  * [x] `editor.destroy()` calls all card teardown callbacks

bantic:
Feedback welcome. Merging this needs to be coordinated with changes to other codebases as described in #235 

checklist:

 * [x] test `unknownCardHandler` usage/defaults for editor

fixes #236


src/js/models/card-node.js
```diff
@@ -54,11 +75,17 @@ export default class CardNode {
...
-      }
+  _validateAndAppendRenderResult(rendered) {
+    if (!rendered) {
+      return;
     }
+
+    let { card: { name } } = this;
+    assert(
+      `Card "${name}" must render dom (render value was: "${rendered}")`,
+      (typeof rendered === 'object')
```
mixonic: perhaps `rendered.nodeType` is a good check for `rendered` being a node.

mixonic:
seems good.

src/js/models/card-node.js
bantic: good call

PR #244, opened 2015-11-23
IE11 Support

mixonic:
This gets tests passing under IE11, and stops using `input` (which is not supported on IE11 contentEditable).

TODO:

* [ ] ~~`input` was triggering a reparse when text content is drag/dropped. Instead of relying on native drag/drop, we should take a cue from other libs and trigger a copy when the user starts dragging. A `drop` can then be a paste event. This is important for us to drag around cards, beyond just restoring old drag behaviors that `input` powered.~~ This behavior is not widely used, and already kind of janky/busted today. Will land this as a followup instead of as part of this commit.
* [x] dropping `input` also breaks text insertion via context menu (like a spelling correction). to restore this functionality, we should look at using `change` events (unknown how viable they are) or DOM mutation events (which are supported in IE11+).

### DIFF
--- a/src/js/utils/cursor.js
+++ b/src/js/utils/cursor.js
@@ -147,8 +147,13 @@ const Cursor = class Cursor {
 
     const range = document.createRange();
     range.setStart(node, offset);
-    this.selection.addRange(range);
-    this.selection.extend(endNode, endOffset);
+    if (direction === DIRECTION.BACKWARD && !!this.selection.extend) {
+      this.selection.addRange(range);
+      this.selection.extend(endNode, endOffset);
+    } else {
+      range.setEnd(endNode, endOffset);
+      this.selection.addRange(range);
+    }
   }
 
   _hasSelection() {
--- a/src/js/utils/lifecycle-callbacks.js
+++ b/src/js/utils/lifecycle-callbacks.js
@@ -17,4 +17,15 @@ export default class LifecycleCallbacksMixin {
     }
     this.callbackQueues[queueName].push(callback);
   }
+  addCallbackOnce(queueName, callback) {
+    if (!queueName) { throw new Error('Must pass queue name to addCallbackOnce'); }
+    let queue = this.callbackQueues[queueName];
+    if (!queue) {
+      this.callbackQueues[queueName] = queue = [];
+    }
+    let index = queue.indexOf(callback);
+    if (index === -1) {
+      queue.push(callback);
+    }
+  }
 }
--- a/src/js/utils/selection-utils.js
+++ b/src/js/utils/selection-utils.js
@@ -11,7 +11,36 @@ function comparePosition(selection) {
 
   const position = anchorNode.compareDocumentPosition(focusNode);
 
-  if (position & Node.DOCUMENT_POSITION_FOLLOWING) {
+  // IE may select return focus and anchor nodes far up the DOM tree instead of
+  // picking the deepest, most specific possible node. For example in
+  //
+  //     <div><span>abc</span><span>def</span></div>
+  //
+  // with a cursor between c and d, IE might say the focusNode is <div> with
+  // an offset of 1. However the anchorNode for a selection might still be
+  // <span> 2 if there was a selection.
+  //
+  // This code walks down the DOM tree until a good comparison of position can be
+  // made.
+  //
+  if (position & Node.DOCUMENT_POSITION_CONTAINS) {
+    return comparePosition({
+      focusNode: focusNode.childNodes[focusOffset],
+      focusOffset: 0,
+      anchorNode, anchorOffset
+    });
+  } else if (position & Node.DOCUMENT_POSITION_CONTAINED_BY) {
+    let offset = anchorOffset - 1;
+    if (offset < 0) {
+      offset = 0;
+    }
+    return comparePosition({
+      anchorNode: anchorNode.childNodes[offset],
+      anchorOffset: 0,
+      focusNode, focusOffset
+    });
+  // The meat of translating anchor and focus nodes to head and tail nodes
+  } else if (position & Node.DOCUMENT_POSITION_FOLLOWING) {
     headNode = anchorNode; tailNode = focusNode;
     headOffset = anchorOffset; tailOffset = focusOffset;
     direction = DIRECTION.FORWARD;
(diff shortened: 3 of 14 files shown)


mixonic:
@bantic ready for a read.

* Replace the `input` event with mutation observer
* Only use `Selection#extend` if needed and present
* Use range/paste based text insertion instead of unsupported
  execCommand('insertText' for tests
* Also fix keyCodes in tests (keyCodes are not charCodes)

Adds a `didReparse` queue for executing text expansions based on the trigger key after the next reparse.

tests/acceptance/cursor-movement-test.js
```diff
@@ -169,57 +170,62 @@ module('Acceptance: Cursor Movement w/ shift', {
   }
 });
 
-test('left arrow when at the end of a card moves the cursor across the card', assert => {
-  let mobiledoc = Helpers.mobiledoc.build(({post, cardSection}) => {
-    return post([
-      cardSection('my-card')
-    ]);
+if (supportsSelectionExtend()) {
```
bantic: For some tests in the past we had a function `skipInPhantom` that ran the tests if they weren't running in phantom.
If there end up being multiple tests that shouldn't run for IE a `skipInIE` function or method on test: `test.skipIf(conditionFn, name, testFn)` would be useful to minimize the noise in diffs.

src/js/utils/cursor.js
```diff
@@ -147,8 +147,13 @@ const Cursor = class Cursor {
 
     const range = document.createRange();
     range.setStart(node, offset);
-    this.selection.addRange(range);
-    this.selection.extend(endNode, endOffset);
+    if (direction === DIRECTION.BACKWARD && !!this.selection.extend) {
```
bantic: Should this use a `supportsSelectionExtend` helper function (lift the one from `tests/helpers`?)

src/js/utils/lifecycle-callbacks.js
```diff
@@ -17,4 +17,15 @@ export default class LifecycleCallbacksMixin {
...
+  addCallbackOnce(queueName, callback) {
+    if (!queueName) { throw new Error('Must pass queue name to addCallbackOnce'); }
+    if (!this.callbackQueues[queueName]) {
+      this.callbackQueues[queueName] = [];
+    }
+    let index = this.callbackQueues[queueName].indexOf(callback);
+    if (index !== -1) {
+      this.callbackQueues[queueName].splice(index, 1);
+    }
+    this.callbackQueues[queueName].push(callback);
```
bantic: Is there a reason to force the callback to always be at the end of the queue like this? If not would be simpler to:
```
if (index === -1) {
  this.callbackQueues[queueName].push(callback);
}
```

src/js/utils/selection-utils.js
```diff
@@ -11,7 +11,23 @@ function comparePosition(selection) {
...
+    return comparePosition({
+      focusNode: focusNode.childNodes[focusOffset],
+      focusOffset: 0,
+      anchorNode, anchorOffset
+    });
+  } else if (position & Node.DOCUMENT_POSITION_CONTAINED_BY) {
+    let offset = anchorOffset - 1;
+    if (offset < 0) {
+      offset = 0;
+    }
```
bantic: `let offset = Math.max(0, anchorOffset - 1);`

What issues are being fixed by the recursive position comparisons?

src/js/utils/cursor.js
mixonic: Hm, I don't think so. The `supportsSelectionExtend` actually does a test when the JS is initially executed, and we don't need that here. I'd like to leave them split apart.

src/js/utils/selection-utils.js
```diff
@@ -20,11 +36,20 @@ function comparePosition(selection) {
...
+      direction = DIRECTION.BACKWARD;
+    } else if (anchorOffset < focusOffset) {
+      headOffset = anchorOffset;
+      tailOffset = focusOffset;
+      direction = DIRECTION.FORWARD;
+    } else {
+      headOffset = anchorOffset;
+      tailOffset = focusOffset;
+      direction = null;
+    }
```
bantic: More of a stylistic suggestion, to me it's more understandable to dry the differences between the 3 scenarios like so:
```
headNode = tailNode = anchorNode;
headOffset = Math.min(anchorOffset, focusOffset); // head always the smaller number
tailOffset = Math.max(anchorOffset, focusOffset); // tail always the larger number
// direction = null if equal, else forward if focus is > offset, backward otherwise
direction = anchorOffset === focusOffset ? null : (focusOffset > anchorOffset ? DIRECTION.FORWARD : DIRECTION.BACKWARD);
```

tests/unit/utils/selection-utils-test.js
```diff
@@ -0,0 +1,83 @@
+const {module, test} = QUnit;
+
+import { comparePosition } from 'mobiledoc-kit/utils/selection-utils';
```
bantic: `import { DIRECTION } from 'mobiledoc-kit/utils/key';`, then can use `DIRECTION.FORWARD`, etc in the tests below

src/js/utils/lifecycle-callbacks.js
mixonic: I felt this was a completeness item, that is should always schedule as late as it was last added. But yeah, where I use it today there is only one thing in the queue regardless. Seem ok to simplify :+1: will do.

bantic:
:+1: The async in tests that use `insertText` is due to the mutation observing firing asynchronously, right?

src/js/utils/selection-utils.js
mixonic: IE picks DOM nodes higher up the tree than Chrome and Safari do. For example in `<div><span>abc</span><span>def</span></div>` with a cursor between `c` and `d` they might say the `focusNode` is `<div>` with an offset of `1`. However the `anchorNode` for a selection might still be `<span>` `2` if there was a selection. We make the upper element descend until a comparison of position can be made.

I can document, there are some test cases added.

tests/unit/utils/selection-utils-test.js
mixonic: :+1:

mixonic:
@bantic correct. I considered passing a callback to be fired on `setTimeout(cb, 0)` or using promises, but wimped out from anything so aggressive. We can ponder more.

src/js/utils/selection-utils.js
bantic: That makes sense. A comment like the above ^ would be helpful. Thanks for adding the tests. I was just too lazy to comprehend the cases they were covering :stuck_out_tongue:

bantic:
Makes sense re the timeouts. Pulling the trigger on promises (or async / await) for tests seems worthwhile at some point, but this looks good to me as is.

PR #248, opened 2015-12-01
Toggle list type sections

### LINKED ISSUE #186: List items can be enabled, removed based on selections
Any list items currently under select should be toggleable.

**A list item under cursor should remove list formatting when toggled**

Content-Kit:

Google Docs:

**Several selected sections should get list formatting when toggled, unless they already have it**

Content-Kit:

Google Docs:

bantic:
Add `postEditor#toggleSection`, `postEditor#setRange`, `editor#selectRange` APIs

`toggleSection` is list-section and list-item aware and will toggle a markup section to/from a list item. Toggling a markup section that is adjacent to a list will join it with that list. Toggling a list item will turn it into a markup section and separate it from the before/after list section. The postEditor run loop will join contiguous list sections of the same type during the before_complete callback queue. This means that deleting a markup section that was between two lists will also cause the lists to be joined together. 

There is some internal refactoring of the cursor movement/selection methods. It is possible that a postEditor method returns a position that has a `section` property that is no longer part of the post abstract after the postEditor's `run()` finishes. This can happen when a markup section is turned into a listSection+listItem and then (in the before_complete queue) merged with an adjacent list.
As a result, it is no longer always safe to use a returned `Position` with a call to one of the cursor movement methods on `editor.cursor` (e.g. moveToPosition/selectRange/selectSection). The `postEditor#setRange` method is added to handle this case.
Mutation methods on the postEditor that could destroy/detach a section can now introspect the current range and update it if they are removing/changing a section that is part of the active `range`. The `postEditor#_joinContiguousListSections` is the only method that does this so far, but eventually all of the methods should be updated to update an active range when necessary. The postEditor automatically renders the updated range at the end of its run loop if `postEditor#setRange` was called.

This is ready for review and merge. If this looks good I will update the documentation to reflect the new APIs shortly before/after merging.

### Possible breaking changes

  * removal of the following methods on `cursor`:
    * `selectSections`, `moveToSection`, `moveToPosition`. None of these was explicitly marked as public.
  * removal of `editor#moveToPosition`. This was not explicitly public before
  * The README mentions `editor#moveToSection`, but this API did not actually exist (it should have been `editor.cursor.moveToSection`). After merge I'll fix up this part of the readme

abc

fixes #186

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -20,6 +20,7 @@ import { setData } from '../utils/element-utils';
 import mixin from '../utils/mixin';
 import EventListenerMixin from '../utils/event-listener';
 import Cursor from '../utils/cursor';
+import Range from '../utils/cursor/range';
 import PostNodeBuilder from '../models/post-node-builder';
 import {
   DEFAULT_TEXT_EXPANSIONS, findExpansion, validateExpansion
@@ -232,14 +233,16 @@ class Editor {
     const range = this.cursor.offsets;
 
     if (this.cursor.hasSelection()) {
-      let nextPosition = this.run(postEditor => postEditor.deleteRange(range));
-      this.cursor.moveToPosition(nextPosition);
-    } else if (event) {
-      const key = Key.fromEvent(event);
-      const nextPosition = this.run(postEditor => {
-        return postEditor.deleteFrom(range.head, key.direction);
+      this.run(postEditor => {
+        let nextPosition = postEditor.deleteRange(range);
+        postEditor.setRange(new Range(nextPosition));
       });
-      this.cursor.moveToPosition(nextPosition);
+    } else if (event) {
+      let key = Key.fromEvent(event);
+      this.run(postEditor => {
+        let nextPosition = postEditor.deleteFrom(range.head, key.direction);
+        postEditor.setRange(new Range(nextPosition));
+      });
     }
   }
 
@@ -248,17 +251,20 @@ class Editor {
 
     event.preventDefault();
 
-    const range = this.cursor.offsets;
-    const cursorSection = this.run(postEditor => {
+    let range = this.cursor.offsets;
+    this.run(postEditor => {
+      let cursorSection;
       if (!range.isCollapsed) {
-        let nextPosition = postEditor.deleteRange(range);
-        if (nextPosition.section.isBlank) {
-          return nextPosition.section;
+        let nextPosition  = postEditor.deleteRange(range);
+        cursorSection = nextPosition.section;
+        if (cursorSection && cursorSection.isBlank) {
+          postEditor.setRange(Range.fromSection(cursorSection));
+          return;
         }
       }
-      return postEditor.splitSection(range.head)[1];
+      cursorSection = postEditor.splitSection(range.head)[1];
+      postEditor.setRange(Range.fromSection(cursorSection));
     });
-    this.cursor.moveToSection(cursorSection);
   }
 
   showPrompt(message, defaultValue, callback) {
@@ -271,25 +277,44 @@ class Editor {
 
   selectSections(sections=[]) {
     if (sections.length) {
-      this.cursor.selectSections(sections);
+      let headSection = sections[0],
+          tailSection = sections[sections.length - 1];
+      this.selectRange(new Range(headSection.headPosition(),
+                                 tailSection.tailPosition()));
     } else {
       this.cursor.clearSelection();
     }
     this._reportSelectionState();
   }
 
-  selectRange(range){
-    this.cursor.selectRange(range);
-    this._reportSelectionState();
+  selectRange(range) {
+    this.range = range;
+    this.renderRange();
   }
 
-  moveToPosition(position) {
-    this.cursor.moveToPosition(position);
+  // @private
+  renderRange() {
+    this.cursor.selectRange(this.range);
     this._reportSelectionState();
+
+    // ensure that the range is "cleaned"/un-cached after
+    // rendering a cursor range
+    this.range = null;
   }
 
   get cursor() {
     return new Cursor(this);
+  }
+
+  // "read" the range from dom unless it has been set explicitly
+  // Any method that sets the range explicitly should ensure that
+  // the range is rendered and cleaned later
+  get range() {
+    return this._range || this.cursor.offsets;
+  }
+
+  set range(newRange) {
+    this._range = newRange;
   }
 
   setPlaceholder(placeholder) {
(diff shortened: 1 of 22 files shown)


src/js/editor/post.js
```diff
@@ -798,13 +801,140 @@ class PostEditor {
...
-    section.tagName = newTagName;
-    this._markDirty(section);
+
+    let sectionsToChange = [];
+    post.walkMarkerableSections(range, section => {
+      sectionsToChange.push(section);
+    });
+    let tagName = everySectionHasTagName ? 'p' : sectionTagName;
+    sectionsToChange.forEach(section => {
+      this.changeSectionTagName(section, tagName);
```
mixonic: I think this can happen in the `walkMarkerableSections` above- the array `sectionsToChange` seems to not serve any logical function

src/js/editor/post.js
```diff
@@ -798,13 +801,140 @@ class PostEditor {
...
+    });
+  }
+
+  _isSameSectionType(section, sectionTagName) {
```
mixonic: Seems like a good method for the `section` itself

src/js/editor/post.js
```diff
@@ -798,13 +801,140 @@ class PostEditor {
...
+  /**
+   * @param {MarkerableSection} section
+   */
+  changeSectionTagName(section, newTagName) {
+    if (isListSectionTagName(newTagName)) {
+      return this._changeSectionToListItem(section, newTagName);
+    } else if (section.isListItem) {
+      return this._changeSectionFromListItem(section, newTagName);
+    } else {
+      section.markers.forEach(m => {
```
mixonic: The branch is hit when the section is not a list item, and when the new tag name is not a list section tag name. So I expect cards is the main use-case?

Regardless I'm unsure why `clearMarkups` is called for each section as a fallback behavior. Can you clarify?

mixonic:
Was there a bug of note in the linked list implementation?

src/js/editor/post.js
bantic: That was the original behavior for `#changeSectionTagName` — I'm not sure what the original rationale is, but that is not something that was added by this PR. I think the idea may have been that markup is not always wanted/correct for some markup section types (like changing a section to an H1 — bold or italicized text may not be wanted in the heading). I think we can do away with that, though — seems fine to keep markers intact when changing between any markerable section.

This method is always called with a markerable section (see the `@param`) — the `else` in this conditional gets hit when it's a non-list-section, non-list-item (aka generic MarkupSection).

src/js/editor/post.js
bantic: agreed

src/js/editor/post.js
bantic: At one point there was an issue with mutation while walking — a section's `next` would get reassigned and the `walkMarkerableSections` callback would skip sections. I will try removing this and see if the tests still pass.

bantic: Removed it and tests are still :green_apple:, seems good

bantic:
> Was there a bug of note in the linked list implementation?

The heuristic for detecting if an item is already in a list doesn't cover every case. In particular, an item can be the only item in another list, in which case its `next` and `prev` will be null, and no error will be thrown. In addition there's a code path that is dead in all circumstances except in this edge case, in which case it (incorrectly) overwrites the tail of the list with the item-from-another-list. This isn't something that we ever encounter in correctly-running code, however, but it does mean the insertion-guard code doesn't protect against this case (and in fact an unusual bug — as described — occurs). I went a little overboard refactoring; I'll separate the LL code into a separate PR for discussion.

bantic:
@mixonic This is ready for another review when you have a chance

src/js/editor/editor.js
```diff
@@ -248,17 +251,20 @@ class Editor {
 
     event.preventDefault();
 
-    const range = this.cursor.offsets;
-    const cursorSection = this.run(postEditor => {
+    let range = this.cursor.offsets;
```
mixonic: So eager to change all these `cursor.offsets` :-D

src/js/editor/editor.js
```diff
@@ -248,17 +251,20 @@ class Editor {
...
+    this.run(postEditor => {
+      let cursorSection;
       if (!range.isCollapsed) {
-        let nextPosition = postEditor.deleteRange(range);
-        if (nextPosition.section.isBlank) {
-          return nextPosition.section;
+        let nextPosition  = postEditor.deleteRange(range);
+        cursorSection = nextPosition.section;
+        if (cursorSection && cursorSection.isBlank) {
```
mixonic: oof, this whole function can probably be refactored to be a lot more clear.

src/js/editor/post.js
```diff
@@ -52,6 +25,21 @@ class PostEditor {
...
+
+  begin() {
+    // mark the editor's range as "dirty"
+    // this forces the editor to cache its range so that
+    // it doesn't try to read from DOM while we are modifying the post abstract
+    this.editor.range = this.editor.cursor.offsets;
+  }
+
+  setRange(range) {
+    // FIXME validate that the range is valid (does not contain marked-for-removal head or tail sections?)
```
mixonic: This is a note to add an assertion? Not really a broken `fixme` item, right?

src/js/utils/cursor/range.js
```diff
@@ -2,19 +2,24 @@ import Position from './position';
...
 export default class Range {
-  constructor(head, tail, direction) {
+  constructor(head, tail=head, direction=DIRECTION.FORWARD) {
     this.head = head;
     this.tail = tail;
     this.direction = direction;
   }
 
-  static create(headSection, headOffset, tailSection, tailOffset) {
+  static create(headSection,             headOffset,
```
mixonic: :-p such whitespace

src/js/utils/linked-list.js
```diff
@@ -107,10 +113,11 @@ export default class LinkedList {
   forEach(callback) {
     let item = this.head;
     let index = 0;
+    let nextItem;
     while (item) {
-      callback(item, index);
-      index++;
-      item = item.next;
+      nextItem = item.next;
+      callback(item, index++);
```
mixonic: Would prefer a `changeTolerantForEach` or something. Or a flag. This change makes `forEach` behave less like it would with an array `for` loop, and thus is a pretty surprising quirk.

We had delete-safe iteration in a few places of the codebase before, we must have.

src/js/editor/editor.js
mixonic: (just in general, doesn't need to be part of this work)

bantic: This was merge-conflicted and I cleaned it up when resolving the merge.

src/js/editor/post.js
bantic: I'll change it to a TODO

PR #250, opened 2015-12-02
Remove styles for the legacy bundled UI

gdub22:
Since this project has pivoted to be programmatic, we no longer need to include the old css


src/css/tooltip.less
```diff
@@ -1,53 +0,0 @@
...
-@-webkit-keyframes fade-in {
-  0% { opacity: 0; }
-  100% { opacity: 1; }
-}
-@keyframes fade-in {
-  0% { opacity: 0; }
-  100% { opacity: 1; }
-}
-
-.__mobiledoc-tooltip {
```
mixonic: Tooltips are still used, actually. The functionality should be extracted to the Ember editor before we drop it here. It will ensure we have appropriate APIs exposed.

src/css/cards.less
```diff
@@ -1,3 +0,0 @@
-.__mobiledoc-card {
-  display: inline-block;
```
mixonic: This is needed to show cards with cursors around them to the left and right.

mixonic:
+1 for dropping Less, and all styles except the card and tooltip stuff.

Agreed that we would like to get rid of all of it soon, but getting this PR merged with the changes possible today seems great progress!

gdub22:
HOLD.  Just realized the demo is installing mobiledoc-kit from npm instead of using the code in the actual repo itself, so this will break some useful styling

src/css/cards.less
zeppelin: Can this be circumvented somehow?

bantic: @zeppelin Are you asking how to make cards so they don't appear as `inline-block`?

zeppelin: @bantic What I mean is to be able to pick which cards are `inline-block`. Currently I can't target individual card containers, which makes impossible to, for example, have two cards, of which one is 100% wide and the other isn't.

PR #251, opened 2015-12-02
Add better guard against inserting item from other list into linked list

bantic:
Adds `__parent` property to items to identify their inclusion in a list.
This allows us to always know when we are inserting an item from another
list. The previous heuristic (checking for presence of `next` and `prev`) is
not correct in all scenarios, leading to unusual bugs.

This refactors `insertBefore` method slightly for clarity, and
simplifies `remove` because we now know for sure when we are removing
an item in this list.

Adds some additional test coverage.

### DIFF
--- a/src/js/utils/linked-list.js
+++ b/src/js/utils/linked-list.js
@@ -1,3 +1,7 @@
+import assert from './assert';
+
+const PARENT_PROP = '__parent';
+
 export default class LinkedList {
   constructor(options) {
     this.head = null;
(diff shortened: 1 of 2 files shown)


src/js/utils/linked-list.js
```diff
@@ -11,9 +15,15 @@ export default class LinkedList {
...
   }
   adoptItem(item) {
+    item[PARENT_PROP]= this;
+    this.length++;
     if (this._adoptItem) { this._adoptItem(item); }
   }
   freeItem(item) {
+    item[PARENT_PROP] = null;
+    item.prev   = null;
+    item.next   = null;
```
mixonic: Shouldn't this already be done by `remove`?

mixonic: Ah you've moved it. As `prev` and `next` are related to position and not ownerships (`adopt`/`free`) IMO the changes should be made in remove.

mixonic:
:+1:

src/js/utils/linked-list.js
bantic: Makes sense. I've moved it. Thanks for looking it over.

PR #254, opened 2015-12-03
Add support for tabs

mixonic:
Tabs are represented by tab characters in Mobiledoc, ala `\t`, and in rendered DOM are represented as an emspace: `\u2003`. This is a compromise for expediency. To use a non-character (a span without contenteditable enabled) we would need to take on many of the challenges atoms area already taking on. I prefer to call this a workable solution and we can tweak the DOM representation at a later time.

Additionally fix an issue with white-space only markers being the first thing in a section. They were disregarded by the renderer, errantly.

demo

This will require followup PRs to the renderers, but is harmless to merge (the renderer will just render a tab character, which in HTML will be invisible).

Fixes https://github.com/bustlelabs/mobiledoc-kit/issues/187

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -36,6 +36,7 @@ import {
   setClipboardCopyData
 } from '../utils/paste-utils';
 import { DIRECTION } from 'mobiledoc-kit/utils/key';
+import { TAB } from 'mobiledoc-kit/utils/characters';
 
 export const EDITOR_ELEMENT_CLASS_NAME = '__mobiledoc-editor';
 
@@ -598,8 +599,7 @@ class Editor {
       this._insertEmptyMarkupSectionAtCursor();
     }
 
-    const key = Key.fromEvent(event);
-
+    let key = Key.fromEvent(event);
     let range, nextPosition;
 
     switch(true) {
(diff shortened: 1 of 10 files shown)


src/js/editor/editor.js
```diff
@@ -598,8 +598,7 @@ class Editor {
       this._insertEmptyMarkupSectionAtCursor();
     }
 
-    const key = Key.fromEvent(event);
-
+    let key = Key.fromEvent(event);
```
bantic: let the `let`ting begin!

src/js/editor/editor.js
```diff
@@ -630,11 +629,24 @@ class Editor {
...
-          let nextPosition = this.run(postEditor => postEditor.deleteRange(range));
+        let { offsets: range } = this.cursor;
+        let { isCollapsed } = range;
+        let nextPosition = range.head;
+        this.run(postEditor => {
+          if (!isCollapsed) {
+            nextPosition = postEditor.deleteRange(range);
+          }
+          if (!range.head.section.isCardSection && key.isTab()) {
+            nextPosition = postEditor.insertText(nextPosition, '\t');
```
bantic: the conditional would read more cleanly (imo) is the `isCardSection` is last (like the `ifCollapsed` conditional a few lines below this):
`if (key.isTab() && !range.head.section.isCardSection)`

Also the `'\t'` should probably be a constant: `TAB_CHARACTER` or similar.

src/js/editor/post.js
```diff
@@ -674,6 +674,25 @@ class PostEditor {
     return this.moveSectionBefore(collection, renderedSection, beforeSection);
   }
 
+  insertText(position, text) {
+    let {marker,offset} = position.markerPosition;
+    let section = position.section;
+    let nextPosition = position;
```
bantic: this could early-return if the section is not markerable, which could improve the clarity:
```
if (!section.isMarkerable) {
  return position;
}
```

src/js/renderers/editor-dom.js
```diff
@@ -43,6 +44,9 @@ function renderHTMLText(marker) {
   text = text.replace(/ ( )/g, () => {
     return ' '+NO_BREAK_SPACE;
   });
+  text = text.replace(/\t/g, () => {
```
bantic: If the editor used/exported its `TAB_CHARACTER` constant ('\t') then this regex could also use it. Seems a bit less brittle than to hardcode "\t" in the two different files.

bantic: The replace function argument is not necessary, also. Not sure if there's any performance difference, but this code is equivalent and (imo) a little more understandable: `text = text.replace(/\t/g, TAB_CHARACTER);`

bantic:
Looks good! @mixonic I added few style and clarity-related suggestions.

mixonic:
:clap:

PR #257, opened 2015-12-08
Document parser hooks

mixonic:
Introduce a public API for parser hooks.

TODO

* [x] Get general :+1: from @bantic 
* [x] Tests
* [x] Refactor to the documented API
* [x] Port existing hook usage here and in TypeSet, smoke test

Fixes https://github.com/bustlelabs/mobiledoc-kit/issues/229

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,47 @@ const expansion = {
 };
 ```
 
+### DOM Parsing hooks
+
+A developer can override the default parsing behavior for leaf DOM nodes in
+pasted HTML.
+
+For example, when an `img` tag is pasted it may be appropriate to
+fetch that image, upload it to an authoritative source, and create a specific
+kind of image card with the new URL in its payload.
+
+A demonstration of this:
+
+```js
+function imageToCardParser(node, builder, {addSection, addMarkerable, nodeFinished}) {
+  if (node.nodeType !== 1 || node.tagName !== 'IMG') {
+    return;
+  }
+  var payload = { src: node.src };
+  var cardSection = builder.createCardSection('my-image', payload);
+  addSection(cardSection);
+  nodeFinished();
+}
+var options = {
+  parserPlugins: [imageToCardParser]
+};
+var editor = new Mobiledoc.Editor(options);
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -199,6 +199,44 @@ const expansion = {
 };
 ```
 
+### DOM Parsing hooks
+
+A developer can override the default parsing behavior for leaf DOM nodes in
+pasted HTML.
+
+For example, when an `img` tag is pasted it may be approprite to
```
bantic: appropriate

README.md
```diff
@@ -199,6 +199,44 @@ const expansion = {
...
+fetch that image, upload it to an authoritative source, and create a specific
+kind of image card with the new URL in its payload.
+
+A demonstraton of this:
```
bantic: demonstration

README.md
```diff
@@ -199,6 +199,44 @@ const expansion = {
...
+```
+
+Parser hooks are called with two arguments:
+
+* `node` - The node of DOM being parsed. This may be a text node or an element.
+* `builder` - The abstract model builder.
+
+If the return value of a parser plugin is truthy, that value is appended
+to the abstract tree. If it is `undefined`, the next parser plugin is
+permitted to attempt parsing. If the value is otherwise fasley (i.e. `null`)
```
bantic: falsey

README.md
```diff
@@ -199,6 +199,44 @@ const expansion = {
...
+then no model is appended to the abract tree and the parser will proceed
```
bantic: abstract :)

bantic:
few typos, but otherwise these docs look :+1: to me. There are some scenarios where a return value may be invalid, however. For instance, if a parser hook returned a marker where a section was expected, I expect the parsing would break later (perhaps much later, at render time) when attempting to deal with a post that had a list of sections that looked like `[markupSection, listSection, marker]`.

Is there a plan to guard against nonsensical (in context) return values for parser hooks?

mixonic:
@bantic perhaps we could assert the correct type is appended to a linked-list during `adoptItem`

Issue #256, opened 2015-12-08
Copy-pasting from Notepad does not work


YoranBrondsema:
Hi,

We found an issue regarding copy-pasting from an external source. On the demo (https://bustlelabs.github.io/mobiledoc-kit/demo/), I can copy-paste text from the page into the editor. Copying from other pages also works. However, when I try to copy-paste from an external source such as the address bar or Notepad, the paste does not work. I tried on both Firefox and Chrome and the issue arises for both browsers.

bantic:
@YoranBrondsema Thanks for reporting this. Can you use [this codepen](http://codepen.io/bantic/pen/wKpOqE) to figure out the exact text/html that is in the clipboard? Please add it as a comment to this issue.

Also a screenshot showing the mobiledoc document and the cursor position before you are attempting to paste.

YoranBrondsema:
@bantic From your Codepen, if `mobiledoc-kit` works the same way, it might be that when you copy non-formatted text from an external source, it doesn't have the data type `text/html`. In that case, `getData('text/html')` is empty. See [my codepen](http://codepen.io/anon/pen/mVyvyJ) (I extended yours to show the result `getData('text/plain')` as well) and the following screenshots.

Copying HTML:
![copying-html](https://cloud.githubusercontent.com/assets/1374412/11669024/1d7f0ffa-9dd9-11e5-9b7e-8c703619d3dd.png)

Copying the URL from the address bar (plain text):
![copying-text](https://cloud.githubusercontent.com/assets/1374412/11669031/2742dcd8-9dd9-11e5-9e17-f0bffbef9c7f.png)

Could this be the bug? It may be because of [this line in paste-utils.js](https://github.com/bustlelabs/mobiledoc-kit/blob/0332d674595c5c4bcdf807d323c357cdd9f77944/src/js/utils/paste-utils.js#L31). If `getData('text/html')` is empty, it maybe should fall back to `getData('text/plain')`. What do you think?

bantic:
@YoranBrondsema Thanks. I'm not sure if this is a bug, exactly, but definitely an oversight. We'll need to update the paste code to handle situations where there is only plaintext available. Thank you for pointing this out. I'll work on this after #249

bantic:
fixed by #262. Follow-up in #263.

PR #262, opened 2015-12-14
Fix issue #256:  Copy-pasting from Notepad does not work

YoranBrondsema:
A PR that addresses https://github.com/bustlelabs/mobiledoc-kit/issues/256. Fallback to `text/plain` if `text/html` is empty.

### DIFF
--- a/src/js/utils/paste-utils.js
+++ b/src/js/utils/paste-utils.js
@@ -30,6 +30,11 @@ export function parsePostFromPaste(pasteEvent, builder, plugins=[]) {
 
   let html = pasteEvent.clipboardData.getData('text/html');
 
+  // Fallback to 'text/plain'
+  if (!html || html.length === 0) {
+    html = pasteEvent.clipboardData.getData('text/plain');
+  }
+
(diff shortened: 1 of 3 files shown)


bantic:
Thanks! As written, this continues to use the HTMLParser to parse the plaintext, so if someone copied the text `"abc <b>bold</b>"` it would get pasted as "abc **bold**". Ideally there would be a text parser in addition to the HTML parser, and it could interpret pasted plaintext into mobiledoc (also splitting into multiple sections by linebreak, for instance).
I'm going to merge this and add an issue to use an actual text parser. If you have interest in tackling that you would be welcome to, otherwise I should get to it sometime this week.
thanks @YoranBrondsema !

YoranBrondsema:
Ah true, hadn't thought of that case. Well at least we have a semi-functional plaintext pasting functionality now so thanks for merging! I'll see what I can do this week, I might find some time to tackle this as well. I'll let you know in https://github.com/bustlelabs/mobiledoc-kit/issues/263 if I start working on it so we don't do double work!

PR #264, opened 2015-12-14
Avoid reparse after initial render

mixonic:
Fixes https://github.com/bustlelabs/mobiledoc-kit/issues/260

### DIFF
--- a/src/js/utils/compiler.js
+++ b/src/js/utils/compiler.js
@@ -1,3 +1,5 @@
+import { forEach } from './array-utils';
+
 export function visit(visitor, node, opcodes) {
   const method = node.type;
   if (!visitor[method]) {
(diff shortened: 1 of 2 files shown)


mixonic:
@bantic some quick eyes here would be great. Works well for me.

src/js/utils/compiler.js
```diff
@@ -9,10 +11,15 @@ export function visit(visitor, node, opcodes) {
...
-    } else {
+    let length = params.length;
+    if (length === 0) {
       compiler[method].call(compiler);
+    } else if (length === 1) {
+      compiler[method].call(compiler, params[0]);
+    } else if (length === 2) {
+      compiler[method].call(compiler, params[0], params[1]);
+    } else {
+      compiler[method].apply(compiler, params);
```
bantic: Curious if this is benchmarked? If so what's the speed-up?

bantic:
:+1: looks good

src/js/utils/compiler.js
mixonic: Nah no serious benchmarks, lifted it from Ember where we use a similar pattern.

PR #265, opened 2015-12-14
Set cursor to collapsed after hitting enter

mixonic:
Range.fromSelection was being used to create a cursor position after hitting enter. The implementation of fromSelection is such that it creates a range and doesn't simply set a collapsed cursor.

Fixes https://github.com/bustlelabs/mobiledoc-kit/issues/261

/cc @bantic


bantic:
:+1:

Issue #259, opened 2015-12-09
Refactor Copy/Paste


bantic:
Copy/Paste behavior is somewhat under-specced in mobiledoc-kit at the moment (and in general — I cannot find a general-purpose algorithm described online for pasting structured content into a document).
This issue attempts to define some declarative behaviors for copy-pasting that align with user expectations. It addresses #249 as well.

## Paste Algorithm for mobiledoc-kit

- pasting when in a blank section
    - replaces that section with the entirety of the pasted post
    - update cursor to end of pasted content
- pasting when cursor is in non-markerable section
    - if cursor is at start of section
        - for each pasted section
            - insert if before the cursor section
            - update cursor to be end of inserted section
    - else
        - for each pasted section
            - insert it after the cursor section
            - update cursor to be end of inserted section
- pasting when cursor is in non-nested markerable (e.g., MarkupSection)
    - if pasted post first leaf section is not markerable
        - if cursor position is at start of section
            - insert each pasted section before the cursor section except the last one
            - if the last one is markerable:
                - prepend its content to the cursor section
                - update cursor position to be the end of the prepended content
            - else
                - insert it before the cursor section
                - update cursor position to be the end of the inserted section
        - else if cursor position is in middle of section
            - split the cursor section into pre- and post-split
            - for each pasted section except the last:
                - insert it before the post-split section
                - update cursor position to be the end of the inserted section
            - if last section is markerable:
                - prepend its content to the cursor section
                - update the cursor position to be the end of the prepended content
            - else
                - insert it  before the post-split section
                - update cursor position to be the end of the inserted section
        - else if cursor is at the end of the section
            - for each pasted section
                - insert it after the cursor section
                - update cursor position to be the end of the inserted section
    - else if pasted post first leaf section is non-nested markerable (e.g. MarkupSection)
        - insert leaf section markers at cursor position
        - update cursor position to be end of inserted content
        - if additional sections:
            - break cursor section at updated cursor position into pre- and post-split
            - for each section except the last:
                - insert it before post-split
                - update cursor position to be end of inserted section
            - for last section
                - if it is markerable:
                    - insert its contents at start of post-split
                    - update cursor position to be end of inserted content
                - else
                    - insert it before post-split
                    - update cursor position to be end of inserted section
    - else if pasted post first leaf section is nested markerable (e.g. ListItem)
        - insert content from 1st item at cursor position
        - update cursor position to be end of inserted content
[...]

PR #267, opened 2015-12-15
Refactor postEditor#insertPost to handle more situations

### LINKED ISSUE #249: Copying/pasting inside list elements doesn't work
I first noticed it in my app where I'm using `Mobiledoc-kit`, but it's easy to reproduce it on http://bustlelabs.github.io/mobiledoc-kit/demo/ as well.

Say I have a paragraph 'X' and a list with a single element 'Y'. I select 'X', copy it, then I select 'Y' and paste over it. That throws an error `Uncaught TypeError: Cannot read property 'renderNode' of undefined` for me on Chrome. Even when the cursor is collapsed, pasting inside `<li>` elements doesn't seem to work and neither does copying, which throws a different error for me: `Uncaught Error: Cannot set section tagName to li`.

I also tried Firefox, and similar thing happens, so the problem seems to be general. I was surprised that nobody has reported it since it seems this problem's been there for a while. :) Do you think there's an easy fix for it or it would require some deeper refactoring?

Thanks!

bantic:
Ready to start getting feedback on this. There are some todos and fixmes I intend to address before this should be merged, though.

  * refactor postEditor#insertPost to handle a larger number of situations
  * fix bug in post#cloneRange when range starts in list item
  * add assert.isPostSimilar, assert.isRenderTreeEqual, assert.isPositionEqual
  * add test for ListSection#clone

Fix #249 #259

### DIFF
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -6,6 +6,7 @@ import mixin from '../utils/mixin';
 import assert from '../utils/assert';
 import { normalizeTagName } from '../utils/dom-utils';
 import Range from '../utils/cursor/range';
+import PostInserter from './post/post-inserter';
 
 function isListSectionTagName(tagName) {
   return tagName === 'ul' || tagName === 'ol';
@@ -606,25 +607,20 @@ class PostEditor {
     if (section.isCardSection) {
       return this._splitCardSection(section, position);
     } else if (section.isListItem) {
-      let [beforeSection, afterSection] = section.splitAtPosition(position);
-      this._coalesceMarkers(beforeSection);
-      this._coalesceMarkers(afterSection);
+      let isLastAndBlank = section.isBlank && !section.next;
+      if (isLastAndBlank) {
+        // if is last, replace the item with a blank markup section
+        let parent = section.parent;
+        let collection = this.editor.post.sections;
+        let blank = this.builder.createMarkupSection();
+        this.removeSection(section);
+        this.insertSectionBefore(collection, blank, parent.next);
 
-      let newSections = [beforeSection, afterSection];
-      let replacementSections = [beforeSection, afterSection];
-
-      if (beforeSection.isBlank && section.isBlank) {
-        const isLastItemInList = section === section.parent.items.tail;
-
-        if (isLastItemInList) {
-          // when hitting enter in a final empty list item, do not insert a new
-          // empty item
-          replacementSections.shift();
-        }
+        return [null, blank];
+      } else {
+        let [pre, post] = this._splitListItem(section, position);
+        return [pre, post];
       }
-
-      this._replaceSection(section, replacementSections);
-      return newSections;
     } else {
       let splitSections = section.splitAtPosition(position);
       splitSections.forEach(s => this._coalesceMarkers(s));
@@ -720,6 +716,29 @@ class PostEditor {
     return this.moveSectionBefore(collection, renderedSection, beforeSection);
   }
 
+  insertMarkers(position, markers) {
+    let { section, offset } = position;
+    assert('Cannot insert markers at non-markerable position',
+           section.isMarkerable);
+
+    let edit = section.splitMarkerAtOffset(offset);
+    edit.removed.forEach(marker => this._scheduleForRemoval(marker));
+
+    let prevMarker = section.markerBeforeOffset(offset);
+    markers.forEach(marker => {
+      section.markers.insertAfter(marker, prevMarker);
+      offset += marker.length;
+      prevMarker = marker;
+    });
+
+    this._coalesceMarkers(section);
+    this._markDirty(section);
+
+    let nextPosition = new Position(position.section, offset);
+    this.setRange(new Range(nextPosition));
+    return nextPosition;
+  }
+
   insertText(position, text) {
     let section = position.section;
     if (!section.isMarkerable) {
@@ -749,7 +768,8 @@ class PostEditor {
 
     let nextNewSection = newSections[0];
     if (nextNewSection.isMarkupSection && section.isListItem) {
-      // put the new section after the ListSection (section.parent) instead of after the ListItem
+      // put the new section after the ListSection (section.parent)
+      // instead of after the ListItem
       collection = section.parent.parent.sections;
       nextSection = section.parent.next;
     }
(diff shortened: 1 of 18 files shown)


src/js/editor/editor.js
```diff
@@ -754,8 +754,7 @@ class Editor {
     let pastedPost = parsePostFromPaste(event, this.builder, this._parserPlugins);
 
     this.run(postEditor => {
-      let nextPosition = postEditor.insertPost(position, pastedPost);
-      postEditor.setRange(new Range(nextPosition));
```
mixonic: Can this be left outside `insertPost`?

We discussed the range stuff in detail but I don't remember if we talked about moving cursor changes into `postEditor` methods. Am I forgetting a reason this must be done?

src/js/editor/post.js
```diff
@@ -720,6 +721,30 @@ class PostEditor {
     return this.moveSectionBefore(collection, renderedSection, beforeSection);
   }
 
+  // TODO write tests for this
```
mixonic: Notable `TODO` ;-)

src/js/editor/post.js
```diff
@@ -720,6 +721,30 @@ class PostEditor {
...
+  insertMarkers(position, markers) {
+    let { section, offset } = position;
+    assert('Cannot insert markers at non-markerable position',
+           section.isMarkerable);
+
+    let edit = section.splitMarkerAtOffset2(offset);
+    edit.removed.forEach(marker => this._scheduleForRemoval(marker));
+    edit.added.forEach(marker => this._markDirty(marker));
```
mixonic: Added things should not need to be marked dirty, they are already dirty. `this._markDirty(section)` as you have below is sufficient.

src/js/editor/post.js
```diff
@@ -720,6 +721,30 @@ class PostEditor {
...
+    markers.forEach(marker => {
+      section.markers.insertAfter(marker, prevMarker);
+      offset += marker.length;
+      prevMarker = marker;
+    });
+
+    this._coalesceMarkers(section);
+    this._markDirty(section);
+
+    position.offset = offset;
```
mixonic: Can this return a new `Position` instead of mutating the one passed in?

```js
return new Position(position.section, offset);
```

src/js/editor/post/post-inserter.js
```diff
@@ -0,0 +1,303 @@
...
+        reference = this.cursorPosition.isHead() ? this.cursorSection :
+                                                   this.cursorSection.next;
+    this.postEditor.insertSectionBefore(collection, blank, reference);
+    this.cursorPosition = blank.tailPosition();
+  }
+
+  _breakMarkerableAtCursor() {
+    let [pre, post] = this.cursorSection.splitAtPosition(this.cursorPosition)
+                          .map(s => s.clone());
+    this.postEditor._coalesceMarkers(pre);
```
mixonic: Maybe `coalesceMarkers` should be public?

src/js/models/_markerable.js
```diff
@@ -11,7 +12,11 @@ export default class Markerable extends Section {
     this.isMarkerable = true;
     this.tagName = tagName;
     this.markers = new LinkedList({
-      adoptItem: m => m.section = m.parent = this,
+      adoptItem: m => {
+        assert(`Cannot insert non-marker into markerable (was: ${m.type})`,
+               m.isMarker);
+        m.section = m.parent = this;
+      },
```
mixonic: This is one of the reasons I dislike the implicit return on `=>` without `{`. Was the return value important here and this refactor is in error? It is difficult to know when reading.

But I know from experience it is not in error. Just a note to share.

src/js/models/_markerable.js
```diff
@@ -127,11 +132,67 @@ export default class Markerable extends Section {
     return edit;
   }
 
+  // splits the marker that contains (or is before) the given offset, at the offset
```
mixonic: period at the end of this line?

src/js/models/_markerable.js
```diff
@@ -127,11 +132,67 @@ export default class Markerable extends Section {
...
+  // differs from `splitMarkerAtOffset` in that it will add an empty marker
+  // when splitting at the start.
+  // This method will modify the section such that `markerBeforeOffset` always
+  // has a valid return value if given the same offset
+  splitMarkerAtOffset2(sectionOffset) {
```
mixonic: Should this be in `postEditor`? Is there a plan to resolve the names?

src/js/models/marker.js
```diff
@@ -116,6 +117,23 @@ const Marker = class Marker extends LinkedItem {
     return markers;
   }
 
+  // @return 2 markers either or both of which could be blank
```
mixonic: To be YUIDoc you need to use `*` quotes:

```js
/**
 * @return {Array} Two markers surrounding the offset. One or both may be blank
 */
```

src/js/models/post.js
```diff
@@ -217,10 +207,27 @@ export default class Post {
     const post = this.builder.createPost();
     const { builder } = this;
 
-    this.walkPostSections(range, section => {
+    let sectionParent = post,
+        listParent = null;
+    this.walkLeafSections(range, section => {
```
mixonic: haha, for a different time but I wonder if this can be structured as an abstract renderer :-)

mixonic:
Some pretty great stuff!!!

src/js/models/_markerable.js
bantic: yeah, I intend to resolve the names

src/js/editor/post/post-inserter.js
bantic: I think `_coalesceMarkers` is an implementation detail that no consumer should have a need to use. It was in use here because this code breaks a section without going through the postEditor. I changed it to use `postEditor#splitSections` instead

src/js/editor/editor.js
bantic: I'm not sure what you mean by "outside", but this can still use the position-returning semantics...I switched it back to the way it was. `postEditor.insertPost` now returns a `Position`

mixonic: Cool. I'm not sure if we had the discussion or not. I know having mutation methods with the side effect of changing cursor position is something we've avoided until now. However the arguments against it are fewer now that we have `setRange`, a safe way to set the position and read it back without the DOM.

bantic:
@mixonic I've added some additional test coverage for new methods and addressed the first round of feedback. This is ready for a final review now. If all looks good, I'll squash and merge.

PR #268, opened 2015-12-16
Add text parser, use it for handling pasted text

### LINKED ISSUE #263: use Text parser for plaintext pasting
#262 adds code to fall back to using the text/plain paste data when text/html is not available, but the plaintext is still parsed into mobiledoc with the HTML parser. There should be a text parser instead that can do things like translate linebreaks into multiple sections, etc.

bantic:
The text parser recognizes multiple sections for text separated by
line breaks and will parse lists when lines of text start with "*"
(unordered) or `"<number>."` (ordered).

fixes #263 

@YoranBrondsema Any feedback you have would be welcome.

### DIFF
--- a/src/js/parsers/text.js
+++ b/src/js/parsers/text.js
@@ -1,0 +1,95 @@
+import assert from 'mobiledoc-kit/utils/assert';
+import {
+  MARKUP_SECTION_TYPE,
+  LIST_SECTION_TYPE
+} from 'mobiledoc-kit/models/types';
+import {
+  DEFAULT_TAG_NAME as DEFAULT_MARKUP_SECTION_TAG_NAME
+} from 'mobiledoc-kit/models/markup-section';
+
+const UL_LI_REGEX = /^\* (.*)$/;
+const OL_LI_REGEX = /^\d\.? (.*)$/;
+const CR = '\r';
+const LF = '\n';
+const CR_REGEX = new RegExp(CR, 'g');
+const CR_LF_REGEX = new RegExp(CR+LF, 'g');
+
+export const SECTION_BREAK = LF;
+
+function normalizeLindEndings(text) {
+  return text.replace(CR_LF_REGEX, LF)
+             .replace(CR_REGEX, LF);
+}
+
+export default class TextParser {
+  constructor(builder, options) {
+    this.builder = builder;
+    this.options = options;
+
+    this.post = this.builder.createPost();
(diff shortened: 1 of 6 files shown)


src/js/parsers/text.js
```diff
@@ -0,0 +1,84 @@
...
+  MARKUP_SECTION_TYPE,
+  LIST_SECTION_TYPE
+} from 'mobiledoc-kit/models/types';
+import {
+  DEFAULT_TAG_NAME as DEFAULT_MARKUP_SECTION_TAG_NAME
+} from 'mobiledoc-kit/models/markup-section';
+
+export const SECTION_BREAK = '\n';
+const UL_LI_REGEX = /^\* (.*)$/;
+const OL_LI_REGEX = /^\d.? (.*)$/;
```
YoranBrondsema: Shouldn't the `.` be escaped? Now I think it will match a digit followed by any character, not sure if that was intended. The line "12 people" would now result in an ordered list with `people`.

src/js/parsers/text.js
```diff
@@ -0,0 +1,84 @@
+import assert from 'mobiledoc-kit/utils/assert';
+import {
+  MARKUP_SECTION_TYPE,
+  LIST_SECTION_TYPE
+} from 'mobiledoc-kit/models/types';
+import {
+  DEFAULT_TAG_NAME as DEFAULT_MARKUP_SECTION_TAG_NAME
+} from 'mobiledoc-kit/models/markup-section';
+
+export const SECTION_BREAK = '\n';
```
YoranBrondsema: Maybe add carriage return `\r` (older versions of Mac OS) and `\r\n` (Windows)?

YoranBrondsema:
> or "." (ordered).

I think you mean "1."?

I just had two small comments regarding the regexes!

src/js/parsers/text.js
bantic: Yes, good catch, thanks!

bantic:
@YoranBrondsema Github escaped my text in the description. It should have said `"<number>. (ordered)"` :)

src/js/parsers/text.js
bantic: Good call, thanks

bantic:
@YoranBrondsema Thanks for reviewing this!

YoranBrondsema:
You're very welcome, thanks for the implementation!

PR #271, opened 2015-12-16
Refactor Position#move to work correctly when prev/next is list section

### LINKED ISSUE #270: cursor movement+selection fails when card borders list section
To reproduce

* create post with card section followed by list section
* position cursor at end of card section
* type `shift+<right-arrow>` to select into the following (list) section
* Error about "markerPositionAtOffset" is thrown

The same error happens when moving left from a card section start into a list section.

bantic:
The `moveRight` and `moveLeft` methods on Position are now smarter about moving to prev/next leaf sections.
Changes most of the card-related cursor movement tests to use `assert.positionIsEqual`.

fixes #270

### DIFF
--- a/src/js/utils/cursor/position.js
+++ b/src/js/utils/cursor/position.js
@@ -76,6 +76,14 @@ const Position = class Position {
     return this.isEqual(this.section.tailPosition());
   }
 
+  /**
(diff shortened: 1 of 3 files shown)


mixonic:
:clap: :clap: :clap: :clap: :clap:

PR #273, opened 2015-12-17
Clear selection in editor#destroy, make editorDomRenderer#destroy safer

bantic:
* Add a `hasRendered` property to Editor and EditorDomRenderer instances.
  * If an editor is destroyed when it has an active cursor, clear that selection.
    This fixes an issue with calling `editor.cursor.offsets` when the
    editor's dom element is selected but the editor has not yet
    rendered. See https://github.com/bustlelabs/ember-mobiledoc-editor/pull/39
  * Update Cursor `hasCursor` and `hasSelection` methods to always
    return false when the editor has not yet rendered.
  * Fix a bug when calling EditorDomRenderer#destroy if it has not yet
    rendered.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -38,6 +38,7 @@ import {
 } from '../utils/paste-utils';
 import { DIRECTION } from 'mobiledoc-kit/utils/key';
 import { TAB } from 'mobiledoc-kit/utils/characters';
+import assert from '../utils/assert';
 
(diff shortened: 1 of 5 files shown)


rlivsey:
:+1:

src/js/editor/editor.js
```diff
@@ -769,6 +772,10 @@ class Editor {
       cardSection.setInitialMode(mode);
     }
   }
+
+  get hasRendered() {
+    return !!this.element;
```
mixonic: technically `element` is present before rendering occurs. seems good for a start though.

PR #275, opened 2015-12-18
marker#deleteValueAtOffset detects if the character is outside the BMP

### LINKED ISSUE #274: deleting characters outside the BMP does not work
Deleting characters outside the [basic multilingual plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) (like [emoji](https://en.wikipedia.org/wiki/Emoticons_(Unicode_block)), [alchemical symbols](https://en.wikipedia.org/wiki/Alchemical_Symbols_(Unicode_block)) and several others) does not work properly because they are composed of two code points. `marker.deleteValueAtOffset` naively splices out only a single "character" from the string.

emoji

bantic:
* [x] acceptance tests for forward- and backward-deleting emoji
  * [x] audit the code for other places that may be affected

There remain some potential places where markers can still be affected by multi-code-point characters. Anywhere in `models/marker.js` that uses the native string-mutation methods (`substr`, e.g.) on `this.value` are affected, but `deleteValueAtOffset` is the only one that will typically trip on this. The other methods will normally be dealing with selections read from the DOM and browsers should not allow a user to select 1/2 of an emoji.

A more comprehensive fix is to write a set of UCS-2 safe string mutation functions and replace every use of built-in string mutation functions in `models/marker.js` with those.

fixes #274

### DIFF
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -515,8 +515,8 @@ class PostEditor {
 
     const offsetToDeleteAt = markerOffset - 1;
 
-    marker.deleteValueAtOffset(offsetToDeleteAt);
-    nextPosition.offset -= 1;
+    let lengthChange = marker.deleteValueAtOffset(offsetToDeleteAt);
+    nextPosition.offset -= lengthChange;
     this._markDirty(marker);
 
     return nextPosition;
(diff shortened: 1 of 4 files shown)


mixonic:
:clap: :clap: :clap: :clap: :clap:

Issue #276, opened 2015-12-22
Copy-pasting from Google Docs and Sheets does not work on Chrome on Windows


YoranBrondsema:
When copying a simple paragraph from Google Docs on a non-Windows Chrome (in my case Linux), the resulting HTML looks like this:

````
<meta http-equiv="content-type" content="text/html; charset=utf-8"><meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-247db770-cad2-1377-b00f-e44a90843b8f"><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:14.666666666666666px;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre-wrap;">simple paragraph</span></p></b><br class="Apple-interchange-newline">
````

This case is handled in mobiledoc-kit. However, when you copy the same thing on the Windows-version of Chrome, you get this:

````
<html><body><!--StartFragment--><meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-4efa44e1-cad3-bc28-33bd-d49e8d9e1f3a"><span style="font-size:14.666666666666666px;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre-wrap;">simple paragraph</span></b><!--EndFragment--></body></html>
````

This causes the DOM parser to fail on the `<!--StartFragment-->` comment, more specifically in the function `normalizeTagName`. There it tries to lower-case the tag name of a Node but Comment nodes have no tag name.

A fix is in the make, you can expect a PR very soon!

PR #277, opened 2015-12-22
Fix bug #276: Copy-pasting from Google Docs and Sheets does not work …

YoranBrondsema:
…on Chrome on Windows

A few remarks:
* When copying from Google Sheets it adds an extraneous span, as you can see in the tests. Not too happy with this but maybe this can be done in a next PR.
* Similarly, when copying two adjacent cells, it glues the cells together (see tests).
* I see you implemented a lot of array utility functions (`forEach`, `filter`, etc...) in `utils/array-utils.js`. Why re-implement those instead of using a library like `lodash`, where you can in a granular way include as NPM modules just the functions you need? Is it because you do operations on `NodeList`, which isn't necessarily supported by Lodash (I don't know the answer to this)?

### DIFF
--- a/src/js/utils/dom-utils.js
+++ b/src/js/utils/dom-utils.js
@@ -1,9 +1,17 @@
 import { forEach } from './array-utils';
 
-const TEXT_NODE_TYPE = 3;
+export const NODE_TYPES = {
+  ELEMENT: 1,
+  TEXT: 3,
+  COMMENT: 8
+};
 
 function isTextNode(node) {
-  return node.nodeType === TEXT_NODE_TYPE;
+  return node.nodeType === NODE_TYPES.TEXT;
+}
+
+function isCommentNode(node) {
+  return node.nodeType === NODE_TYPES.COMMENT;
 }
 
 // perform a pre-order tree traversal of the dom, calling `callbackFn(node)`
@@ -112,6 +120,7 @@ export {
   addClassName,
   normalizeTagName,
   isTextNode,
+  isCommentNode,
   parseHTML,
   findOffsetInElement
 };
--- a/tests/unit/parsers/html-google-sheets-test.js
+++ b/tests/unit/parsers/html-google-sheets-test.js
@@ -1,0 +1,86 @@
+import HTMLParser from 'mobiledoc-kit/parsers/html';
+import PostNodeBuilder from 'mobiledoc-kit/models/post-node-builder';
+import Helpers from '../../test-helpers';
+
+const {module, test} = Helpers;
+
+let parser;
+
+module('Unit: Parser: HTMLParser Google Sheets', {
+  beforeEach() {
+    const options = {};
+    const builder = new PostNodeBuilder();
+    parser = new HTMLParser(builder, options);
+  },
+  afterEach() {
+    parser = null;
+  }
+});
+
+// No formatting
+test('#parse returns a markup section when given a cell without formatting', (assert) => {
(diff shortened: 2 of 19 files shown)


mixonic:
> Is it because you do operations on NodeList, which isn't necessarily supported by Lodash

Close, we want to transparently use these utils on the linked list construct used in our models. The fallback is pretty simple, we just check for an implementation on the object before falling back to an array implementation.

There are things in the codebase way more in need of attention than that :-)

Can you take a look at simply adding the `isCommentNode` check to [_isSkippable](https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/parsers/section.js#L302)? This seems like it is where similar work is already done.

src/js/utils/dom-utils.js
```diff
@@ -71,7 +76,11 @@ function addClassName(element, className) {
 }
 
 function normalizeTagName(tagName) {
-  return tagName.toLowerCase();
+  if (!tagName) {
```
mixonic: And this can probably be reverted if you move the fix as suggested in my comment. IMO this function should not fail silently as you changed it to: If there is no `tagName` to be normalized our code is definitely not doing what we expected it to do.

tests/unit/parsers/html-google-sheets-test.js
```diff
@@ -0,0 +1,77 @@
...
+import { CARD_TYPE } from 'mobiledoc-kit/models/types';
+
+const {module, test} = Helpers;
+
+function parseHTML(html, options={}) {
+  let builder = new PostNodeBuilder();
+  return new HTMLParser(builder, options).parse(html);
+}
+
+module('Unit: Parser: HTMLParser Google Sheets');
```
mixonic: I'd prefer to see these without the big fixtures file. The best for of testing would also not be to test against another parsed string, but against a post. For example:

```js
test('#parse returns a markup section when given single line of text', (assert) => {
  let text = 'some text';
  let post = parser.parse(text);
  let expected = Helpers.postAbstract.build(({post, markupSection, marker}) => {
    return post([markupSection('p', [marker('some text')])]);
  });

  assert.postIsSimilar(post, expected);
});
```

Taken from https://github.com/bustlelabs/mobiledoc-kit/blob/e003067ec9a49ceebd7520b5c785f91136e3261c/tests/unit/parsers/text-test.js#L26-L34

This testing is more exact, only the actual values are parsed. The expected value is just an abstract.

mixonic:
Left a few comments @YoranBrondsema! Thanks for this :-)

YoranBrondsema:
Thanks for the review. I addressed the code part comments, I will address the test part tonight or tomorrow.

YoranBrondsema:
@mixonic alright did the changes to both the code and the tests. Let me know!

mixonic:
This is great @YoranBrondsema. Can you rebase it into a single commit and I'll merge?

YoranBrondsema:
@mixonic I was having some trouble rebasing completely, I think I messed up earlier. Is it OK to leave it like this? It has one extra merge with master which I can't seem to get rid off (I'm not a Git expert :/).

mixonic:
@YoranBrondsema it rebased cleanly for me in https://github.com/bustlelabs/mobiledoc-kit/pull/281, closing this via that PR. Thank you for this fix and the excellent tests!

YoranBrondsema:
Cool, thanks for the rebase and the merge!

PR #291, opened 2016-01-11
Always `setRange` in `toggleMarkup` and `toggleSection`

### LINKED ISSUE #285: when editor element is blurred but has selection, still possible to insert text
When the `document.activeElement` is an element outside the editor's element but `window.getSelection()` is within the editor's element, subsequent keystrokes will refocus onto the editor element and insert text, but they will *not fire event handlers* that were registered on the element. This causes issues related to inserting text in places it shouldn't go, such as around a card.

To reproduce in the demo:

  * insert an image card
  * position cursor at end of the card
  * click one of the section-changing buttons (e.g. "headline")
  * type text
  * the element's keydown handler will not get triggered (because the document's `activeElement` is the button that was clicked), but since the selection is still in the (contenteditable) editor element, the text will be inserted:

bug

### LINKED ISSUE #287: "section.splitMarkerAtOffset is not a function" when card is selected and `toggleMarker` called
When a card is selected, `postEditor#toggleMarkup` will call `splitMarkerAtOffset` on the card section, which results in an undefined function error.

To reproduce in the demo:

  * insert a card (click "add image")
  * select the card
  * click "bold"

bug

bantic:
This helps prevent a situation where the editor element has a selection
but it is not the active element. Typing while another element (like a
button) is focused (active) but the editor element has the selection is
problematic: The browser refocuses on the editor element and starts
inserting text, but it handles the input before it has focused, which
means the key* handlers do not fire.

  * renamed Position#emptyPosition and Range#emptyRange ->
blankPosition, blankRange. Adds `isBlank` property to Position and
Range.
  * Defensively avoid attempting to render a cursor when the range is blank.
  * Add `splitMarkerAtOffset` to Section base class (fixes #287)

cc @mixonic 

Fixes #285

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -306,7 +306,11 @@ class Editor {
 
   // @private
   renderRange() {
-    this.cursor.selectRange(this.range);
+    if (this.range.isBlank) {
+      this.cursor.clearSelection();
+    } else {
+      this.cursor.selectRange(this.range);
+    }
     this._reportSelectionState();
 
     // ensure that the range is "cleaned"/un-cached after
--- a/src/js/utils/cursor/position.js
+++ b/src/js/utils/cursor/position.js
@@ -37,16 +37,17 @@ const Position = class Position {
   constructor(section, offset=0) {
     this.section = section;
     this.offset = offset;
+    this.isBlank = false;
   }
 
-  static emptyPosition() {
+  static blankPosition() {
     return {
       section: null,
       offset: 0,
       marker: null,
       offsetInTextNode: 0,
-      _isEmpty: true,
-      isEqual(other) { return other._isEmpty; },
+      isBlank: true,
+      isEqual(other) { return other.isBlank; },
       markerPosition: {}
     };
   }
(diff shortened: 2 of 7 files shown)


src/js/utils/cursor/position.js
```diff
@@ -63,6 +63,10 @@ const Position = class Position {
     return this.markerPosition.offset;
   }
 
+  get isBlank() {
+    return false;
```
mixonic: Why not a property set inside the `constructor`?

src/js/editor/editor.js
```diff
@@ -306,7 +306,9 @@ class Editor {
 
   // @private
   renderRange() {
-    this.cursor.selectRange(this.range);
+    if (!this.range.isBlank) {
+      this.cursor.selectRange(this.range);
```
mixonic: If the range is blank, I would expect the selection to be cleared. Ignoring the set range seems incorrect.

src/js/utils/cursor/position.js
bantic: No particular reason; I can change it. FWIW I don't think that defining a getter using `get X()` versus setting it on `this` in the constructor changes the shape of the object, in case that is a concern. I'll put it in the constructor for clarity, though.

src/js/editor/editor.js
bantic: Sure, that makes sense to me, too. I'll change it.

PR #293, opened 2016-01-13
Replace `throw Error` usage with `assert`, use custom `MobiledocError`

bantic:
This allows upstream consumers to detect and handle MobiledocKit errors
specially if needed.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -74,9 +74,8 @@ const CALLBACK_QUEUES = {
  */
 class Editor {
   constructor(options={}) {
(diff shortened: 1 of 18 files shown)


mixonic:
> This allows upstream consumers to detect and handle MobiledocKit errors
specially if needed.

Does it? If you want consumers to detect and handle errors in special ways, we need a public API for replacing the assert logic. That means documentation etc as well. I don't think we have any of that.

Looks good though.

mixonic:
```
assert(msg, false)
```

is used a lot here (explicit `false`) and seems unfortunate. Meh.

PR #294, opened 2016-01-14
Handle spacebar keydown

### LINKED ISSUE #292: typing "<space><space>" in empty paragraph removes the whitespace
On Chrome, typing two spaces in a row in an empty paragraph causes them to both be removed and the paragraph is rerendered as an empty paragraph.

bug

bantic:
Captures keyDown event for `<space>` key and handles it semantically, preventing the event default and inserting a space character into the post. This obviates some mutation-observer code. Browsers fire multiple mutations (removing all spaces in the text node, replacing with spaces + an `&nbsp;`) when typing spaces, and there were issues handling the space appropriately when those mutation observers did not fire at the expected intervals (which could happen if, e.g., there were additional mutation listeners added to the window).

Refactors text expansions to happen on `keyDown` of the trigger key rather than using `setTimeout` and the `DID_REPARSE` callback queue. This fixes another issue where typing characters too quickly would not cause the text expansion to fire (typing "* X" very quickly, e.g., could make it so that the text expansion code ran after the "X" was inserted and would not properly expand into a list item).

cc @mixonic 

fixes #292

### DIFF
--- a/src/js/renderers/editor-dom.js
+++ b/src/js/renderers/editor-dom.js
@@ -28,22 +28,36 @@ function createElementFromMarkup(doc, markup) {
   return element;
 }
 
+const TWO_SPACES         = `${SPACE}${SPACE}`;
+const SPACE_AND_NO_BREAK = `${SPACE}${NO_BREAK_SPACE}`;
+const SPACES_REGEX       = new RegExp(TWO_SPACES, 'g');
+const TAB_REGEX          = new RegExp(TAB, 'g');
+const endsWithSpace = function(text) {
+  return endsWith(text, SPACE);
+};
+const startsWithSpace = function(text) {
+  return startsWith(text, SPACE);
+};
+
 // FIXME: This can be done more efficiently with a single pass
 // building a correct string based on the original.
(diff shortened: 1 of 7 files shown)


src/js/renderers/editor-dom.js
```diff
@@ -28,22 +28,36 @@ function createElementFromMarkup(doc, markup) {
...
+  text = text.replace(SPACES_REGEX, SPACE_AND_NO_BREAK)
+             .replace(TAB_REGEX,    TAB_CHARACTER);
+
   // If the first marker has a leading space or the last marker has a
   // trailing space, the browser will collapse the space when we position
   // the cursor.
   // See https://github.com/bustlelabs/mobiledoc-kit/issues/68
   //   and https://github.com/bustlelabs/mobiledoc-kit/issues/75
-  if (!marker.next && endsWith(text, SPACE)) {
+  if (endsWithSpace(text) && !marker.next) {
```
mixonic: the `!marker.next` check is probably faster and should be first, but meh.

mixonic:
:clap: :clap: :clap:

PR #303, opened 2016-01-25
Satisfy newer jshint

ef4:
Newish versions of jshint (I'm using 2.9.1) have bugfixes that cause it to discover failures that were missed by earlier versions.

 - I had to remove the `es3` option because it doesn't not work correctly alongside `esnext`
 - I created a new scope in editor.js because a switch case was shadowing outer variables
 - W083 is the "don't make functions within a loop" error, which is sensible in general, but not needed when you're sure that the functions in question will not outlive a single loop iteration.


mixonic:
:clap: thanks @ef4

PR #304, opened 2016-01-25
Report active markups even when selection is collapsed

ef4:
This makes `markupsInSelection` report active markups even when the selection is collapsed.

If you run the demo, you can see why this is helpful:

 - mark some words bold
 - move the cursor in and out of the bold region
 - observe that the toolbar button for "Bold" correctly shows active state when you're inside the bold region. Prior to this change, the toolbar button would only show active state if you highlighted one or more characters inside a bold region.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -381,7 +381,7 @@ class Editor {
   }
 
   get markupsInSelection() {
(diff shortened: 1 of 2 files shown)


bantic:
This is excellent, thank you!

PR #305, opened 2016-01-26
Refactor cursor positioning

bantic:
It was previously possible to create a position with a null section or
an invalid section (like a list section). A position must have a section
that the cursor can be rendered at; this PR adds an assertion to that
fact.

Before this change, if the window.getSelection() pointed at the editor's
element, Position#fromNode would naively assume that the post's first
section at offset 0 was the position, but this creates an invalid
position when the first section is a list section.

This tightens up some of the cursor positioning and adds a bunch of tests for `Position#fromNode` in various scenarios.

 * add Post#headPosition and #tailPosition
 * add Section#isSection property
 * Refactor Position#fromNode
 * tests for Position#fromNode for text nodes, element nodes, card-related nodes
 * Position constructor asserts it has a valid leaft section and offset

### DIFF
--- a/src/js/models/_section.js
+++ b/src/js/models/_section.js
@@ -14,6 +14,8 @@ export default class Section extends LinkedItem {
     this.type = type;
     this.isMarkerable = false;
     this.isNested = false;
+    this.isSection = true;
+    this.isLeafSection = true;
   }
 
   set tagName(val) {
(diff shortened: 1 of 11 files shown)


mixonic:
:tada: :tada: :tada:

PR #307, opened 2016-01-28
Add MutationHandler, reparse entire post when new nodes appear

### LINKED ISSUE #300: Dragging text to another mobiledoc-editor causes a `Could not find parent section for un-mapped text node` error
When I try to move text from one mobiledoc-editor to another, it throws an assertion error at this line: https://github.com/bustlelabs/mobiledoc-kit/blob/3a22d87869a439748b02379ed8bd49559cee7c50/src/js/utils/cursor/position.js#L148

I'm looking into it now but if you have an idea then let me know!

Yoran

bantic:
## summary of changes

  * adds editor#_reparseSections and editor#_reparsePost
  * adds RenderTree#isDirty, use it to determine whether to rerender
    the cursor position after reparsing
  * tests for changing text and element nodes in the editor dom
  * Remove editor#reparse and #_reparseCurrentSection

## previous reparsing algorithm
Previously, the reparsing algorithm was, roughly:
  * after detecting a mutation(s) (any mutation: added/removed node or any character data change),
  * call editor#_reparseCurrentSection, which:
  * reparse only the section that currently has the cursor in it

This misses changes to the DOM that happened outside the section with the cursor (which happens when dragging in text to the editor if it has no cursor (ie is blank), and/or when dragging in text that causes the browser to insert dom outside the section that has the cursor (ie when dragging multiple lines of text)).

## new reparsing algorithm

Now, the reparsing algorithm inspects the mutation events to determine which section(s) are affected or if the entire post is affected and must be reparsed. The new algorithm is, roughly:
  * for each mutation event, find the significant nodes (added nodes are significant, the target element from which nodes are removed is significant, and the text nodes that have changed character data are significant)
  * prune those significant nodes by removing any that are not attached to any parent node (this can happen when the browser makes multiple mutations at once; when the mutation observer callback fires some of the mutations now point at dom nodes that have since been removed)
  * for each remaining significant node, look up its section from the render tree. If no section is found, we must reparse the entire post (because a node must have been added in the editor outside of one of our rendered sections)
  * either reparse the entire post (if necessary), or reparse the section(s) that we found in the previous step
  * if reparsing sections, and any section gets marked dirty, we must rerender the cursor

Changes to DOMParser#reparseSection:
  * the reparseSection method was primarily intended to detect text changes due to character input to the text node(s) in a section. As such, it assumes the DOM in that section is "clean" and it only walks the text nodes and fixes up their corresponding markers so that the post model represents what is in DOM and its render nodes are marked clean. (A better method name might be `readSectionFromDOM`).
  * but it is possible to change the DOM nodes in a section in other ways, particularly by dragging in text. In this case the browser may insert new elements (spans with inline styles, in particular) for the dragged-in text, and it's no longer safe to assume that the section's DOM is "clean". We want to read the DOM for that section but then also rerender it (to remove the extraneous DOM nodes that the browser may have added)
  * to facilitate this, the DOMParser#reparseSection method is changed so that it will update-in-place a marker when it encounters a known text node, but any *new* text nodes (i.e. ones that have correspond to no render node) will cause it to mark the section's render node as dirty. The editor, after reparsing the section, checks to see if the render tree is dirty and if so it rerenders (and also rerenders the cursor position)

fixes #300

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -39,6 +39,7 @@ import {
 import { DIRECTION } from 'mobiledoc-kit/utils/key';
 import { TAB, SPACE } from 'mobiledoc-kit/utils/characters';
 import assert from '../utils/assert';
+import MutationHandler from 'mobiledoc-kit/editor/mutation-handler';
 
 export const EDITOR_ELEMENT_CLASS_NAME = '__mobiledoc-editor';
 
@@ -89,10 +90,6 @@ class Editor {
     DEFAULT_TEXT_EXPANSIONS.forEach(e => this.registerExpansion(e));
     DEFAULT_KEY_COMMANDS.forEach(kc => this.registerKeyCommand(kc));
 
-    this._mutationObserver = new MutationObserver(() => {
-      this.handleInput();
-    });
-    this._isMutationObserved = false;
     this._parser   = new DOMParser(this.builder);
     this._renderer = new Renderer(this, this.cards, this.unknownCardHandler, this.cardOptions);
 
@@ -137,9 +134,9 @@ class Editor {
     }
 
     this.runCallbacks(CALLBACK_QUEUES.WILL_RENDER);
-    this.removeMutationObserver();
-    this._renderer.render(this._renderTree);
-    this.ensureMutationObserver();
+    this._mutationHandler.suspendObservation(() => {
+      this._renderer.render(this._renderTree);
+    });
     this.runCallbacks(CALLBACK_QUEUES.DID_RENDER);
   }
 
@@ -154,6 +151,8 @@ class Editor {
     clearChildNodes(element);
 
     this.element = element;
+    this._mutationHandler = new MutationHandler(this);
+    this._mutationHandler.startObserving();
 
     if (this.isEditable === null) {
       this.enableEditing();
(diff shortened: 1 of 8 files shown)


mixonic:
@bantic can you create a followup bug issue related to the dom subtree listener? And perhaps one regarding the "just change marker value if only text changed" optimization?

src/js/editor/editor.js
```diff
@@ -565,6 +540,16 @@ class Editor {
     this.addCallback(CALLBACK_QUEUES.CURSOR_DID_CHANGE, callback);
   }
 
+  /**
+   * @method didReparse
+   * @param {Function} callback This callback is called after any part of the
+   *        post is reparsed
+   * @public
+   */
+  didReparse(callback) {
```
mixonic: This should also be documented in [the README](https://github.com/bustlelabs/mobiledoc-kit#editor-lifecycle-hooks)

src/js/editor/mutation-handler.js
```diff
@@ -0,0 +1,129 @@
...
+   *     * if character data changed
+   *       * target node is the mutation event's target (text node)
+   *     * filter out nodes that are no longer attached (parentNode is null)
+   *   * for each remaining node:
+   *   *  find its section, add to sections-to-reparse
+   *   *  if no section, reparse all (and break)
+   */
+  _handleMutations(mutations) {
+    let reparsePost = false;
+    let sections = new Set();
```
mixonic: I think `Set` does not transpile. It must be polyfiled. Support looks iffy (but partially there) for IE11, Node 0.12 (FastBoot target), Safari and iOS 8.

src/js/models/render-tree.js
```diff
@@ -12,6 +12,12 @@ export default class RenderTree {
   get rootNode() {
     return this._rootNode;
   }
+  /**
+   * @return {Boolean}
+   */
+  get isDirty() {
```
mixonic: This seems a bit funky, but it make sense with our current algo. In theory a branch of the tree could be dirty and the root node could not be dirty, and this logic would not catch that case.

Practically, today we always walk and dirty parent nodes anyway.

Basically I have a mild concern this would make a later refactoring of the render-tree complicated by implying dirty semantics based on the root node.

src/js/editor/mutation-handler.js
mixonic: A nm this is our own Set polyfill already :-)

mixonic:
Really great.

bantic:
after discussion, removing `editor#didReparse` hook, since this callback queue was not used for anything before, and there is currently no known use case for it.

YoranBrondsema:
@bantic Sorry I haven't had the opportunity to go over it. Good job, thanks!

bantic:
@YoranBrondsema no worries. This is released in v0.8.0-beta.4 of mobiledoc-kit, fyi, please give it a try if you can and let me know if you encounter any problems.

YoranBrondsema:
@bantic Cool will do!

PR #309, opened 2016-02-01
Use marker's rendered `markupElement` for destroying marker render nodes

### LINKED ISSUE #299: Formatting text deletes it
We’ve had several reports over time of formatting (bold, italic, or links) removing the contents instead. Today and yesterday (on `0.8.0-beta.3`) from two users:

> “tried italicizing in this post [...] and it deleted the text”

> “tried bolding in this piece: [...] and it just deleted the entire text
> i refreshed the page and now it’s there, but every time i try to bold, it deletes
> i even tried clicking the “B” instead, and still happens
> mac, chrome, and was trying to bold the next at the top, but seems once i added a space between the existing text and what i was trying to add, i could bold without an issue
> but for some reason whenever i tried to do it in the text it was erasing the text in its entirety”

### LINKED ISSUE #306: rerendering bold+plaintext+bold markers causes error: "Failed to execute 'replaceChild' on ‘Node’"
To reproduce:

  * type 3 characters of bold text: **bab**
  * select middle character and un-bold it (using cmd-B or "bold" button): **b**a**b**
  * re-bold the middle character -> `Uncaught NotFoundError: Failed to execute 'replaceChild' on 'Node': The node to be replaced is not a child of this node.`

The error comes from the editor-dom's attempt to [reattach the rendered markup section](https://github.com/bustlelabs/mobiledoc-kit/blob/af5ae91300abcfd57645b45a3a1d3ec3ce44f491/src/js/renderers/editor-dom.js#L155).

bantic:
Changes the editor-dom renderer to store a reference to the last markup element it created (as `renderNode.markupElement`) when rendering a marker. Use this markup element in the destroy hook for markers rather counting the marker's markups and walking the DOM.

Previously, when destroying a marker, the renderer would count the marker's markups and walk up that many `parentNode`s from the renderNode's `element` (the textNode for the marker). But it is possible for a marker's number of markups to change after it has been rendered but before it is removed. In these cases, walking up `marker.markups.length` of `parentNode`s from the marker's textNode `element` could result in incorrectly removing more DOM (i.e. a parent P element) than it should.

Fixes #306
Fixes #299

### DIFF
--- a/src/js/models/render-node.js
+++ b/src/js/models/render-node.js
@@ -13,6 +13,9 @@ export default class RenderNode extends LinkedItem {
     this._childNodes = null;
     this._element = null;
     this.renderTree = renderTree;
+
+    // RenderNodes for Markers keep track of their markupElement
+    this.markupElement = null;
   }
   isAttached() {
     assert('Cannot check if a renderNode is attached without an element.',
--- a/src/js/renderers/editor-dom.js
+++ b/src/js/renderers/editor-dom.js
@@ -22,7 +22,7 @@ export const SPACE = ' ';
 export const ZWNJ = '\u200c';
 
 function createElementFromMarkup(doc, markup) {
-  var element = doc.createElement(markup.tagName);
+  let element = doc.createElement(markup.tagName);
   Object.keys(markup.attributes).forEach(k => {
     element.setAttribute(k, markup.attributes[k]);
   });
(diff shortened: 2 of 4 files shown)


bantic:
@mixonic Any feedback?

mixonic:
:+1:

PR #311, opened 2016-02-02
Atoms rebased unsquashed

bantic:
rebased/squashed/unsquashed version of #226, with tests for the cases mentioned in https://github.com/bustlelabs/mobiledoc-kit/pull/226#issuecomment-173630690

**Ready for review**

cc @mixonic @rlivsey

### DIFF
--- a/ATOMS.md
+++ b/ATOMS.md
@@ -1,0 +1,58 @@
+# Mobiledoc Atoms
+
+Atoms are effectively read-only inline cards.
+
+## Atom format
+
+An atom is a javascript object with 3 *required* properties:
+
+  * `name` [string] - The name of this atom in the mobiledoc
+  * `type` [string] - The output of this atom. Valid values are 'dom', 'html', and 'text'
+  * `render` [function] - Invoked by the renderer to render this atom
+
+## Atom rendering
+
+The `render` function on an atom is called by an instance of a renderer and passed an object with the following four properties:
+
+  * `env` [object] - A set of environment-specific properties
+  * `options` [object] - Rendering options that were passed to the renderer (as `cardOptions`) when it was instantiated
+  * `payload` [object] - The data payload for this atom from the mobiledoc
+  * `value` [string] - The textual representation to for this atom
+
+The return value of the `render` function will be inserted by the renderer into the rendered mobiledoc.
+The return value can be null if an atom does not have any output. If there is a return value it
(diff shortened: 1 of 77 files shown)


rlivsey:
:+1: awesome stuff

src/js/editor/key-commands.js
```diff
@@ -53,6 +53,26 @@ export const DEFAULT_KEY_COMMANDS = [{
...
+  run(editor) {
+    let range = editor.cursor.offsets;
+    let {head: {section}} = range;
+    editor.run(postEditor => {
+      postEditor.setRange(new Range(section.headPosition()));
+    });
+  }
+}, {
+  // FIXME restrict to OS X only?
+  str: 'CTRL+E',
```
mixonic: This is a new feature? Or something that broke?

src/js/models/image.js
```diff
@@ -12,11 +11,7 @@ export default class Image extends Section {
     return false;
   }
 
-  headPosition() {
-    return new Position(this, 0);
-  }
-
-  tailPosition() {
-    return new Position(this, 1);
+  get length() {
```
mixonic: IMO this whole model should go the way of the Dodo before 1.0. images are just cards.

src/js/editor/key-commands.js
bantic: The ctr-A and ctr-E OS X native movement commands don't work when there are atoms in the text (the cursor focus gets moved to the inside of the atom). I have a branch will PR after this that addresses these FIXMEs and scopes to Mac only.

src/js/models/image.js
bantic: 100% agree

ATOMS.md
```diff
@@ -0,0 +1,58 @@
...
+
+The return value of the `render` function will be inserted by the renderer into the rendered mobiledoc.
+The return value can be null if an atom does not have any output. If there is a return value it
+must be of the correct type (a DOM Node for the dom renderer, a string of html or text for the html or text renderers, respectively).
+
+#### `env`
+
+`env` always has the following properties:
+
+  * `name` [string] - the name of the card
```
praxxis: name of the card or atom? Atoms are said to be effectively cards, but I don't read that as "atom and card are interchangeable". We also use `card` as the var name in the example on line 49

bantic: Thanks @praxxis, you're right. Will fix this.

PR #318, opened 2016-02-09
WIP Undo redo

bantic:
*do not merge*

Beginning work to enable a naive version of undo/redo in Mobiledoc-Kit. This stores a snapshot (including the cursor position when possible) when changing the mobiledoc (specifically in `editor.run`).

A few things to do:
  * [x] basic tests for undoability
  * [x] limit the undo stack to reduce memory pressure
  * [ ] look at and test edge cases relating to the post (list items, cards, atoms)
  * [ ] test dragging text in and hitting undo (change the textContent of a node, setTimeout to allow the parser, then assert) 
  * [x] look at and test edge cases for positioning (when the position cannot be determined, i.e., editor does not have focus and user clicks a button to add a card)
  * [x] look at and test edge cases related to mutation handling (dragging in a single word or multiple sections)
  * [x] implement a redo stack
  * [x] allow configuring the length of the undo/redo stack

### DIFF
--- a/src/js/editor/edit-history.js
+++ b/src/js/editor/edit-history.js
@@ -1,0 +1,119 @@
+import mobiledocParsers from 'mobiledoc-kit/parsers/mobiledoc';
+import Range from 'mobiledoc-kit/utils/cursor/range';
+import Position from 'mobiledoc-kit/utils/cursor/position';
+import FixedQueue from 'mobiledoc-kit/utils/fixed-queue';
+
+function findLeafSectionAtIndex(post, index) {
+  let section;
+  post.walkAllLeafSections((_section, _index) => {
+    if (index === _index) {
+      section = _section;
+    }
+  });
+  return section;
+}
+
+export class Snapshot {
+  constructor(editor) {
+    this.mobiledoc = editor.serialize();
+    this.editor = editor;
+
+    this.snapshotRange();
+  }
+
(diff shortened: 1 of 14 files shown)


src/js/editor/edit-history.js
```diff
@@ -0,0 +1,123 @@
...
+
+  stepBackward(postEditor) {
+    console.log('stepping backward');
+    // Throw away the pending snapshot
+    this._pendingSnapshot = null;
+
+    let snapshot = this._undoStack.pop();
+    if (snapshot) {
+      console.log('pushing onto redo stack');
+      this._redoStack.push(new Snapshot(this.editor));
```
mixonic: Can you just push the `snapshot` to the redo stack?

src/js/editor/edit-history.js
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      console.log('no snapshot to step backward to');
+    }
+  }
+
+  stepForward(postEditor) {
+    console.log('stepping forward');
+    let snapshot = this._redoStack.pop();
+    if (snapshot) {
+      this._restoreFromSnapshot(snapshot, postEditor);
```
mixonic: Seems like the `snapshot` should also be pushed on to the undo stack. Supporting `undo, undo, redo, undo, redo` for example.

PR #327, opened 2016-02-15
Replace broken `Node.contains` with custom `containsNode`

### LINKED ISSUE #325: 0.8.3 is broken on IE11
See below. I haven't tried with > 0.8.3 but from the changelog I don't expect it to work there. I am looking into this bug now but any ideas are welcome.

ie11

YoranBrondsema:
Fixes #325.

On IE, `Node.contains` is only defined for Elements, not Nodes (see https://connect.microsoft.com/IE/feedback/details/780874/node-contains-is-incorrect).


bantic:
thanks @YoranBrondsema! This looks great

PR #340, opened 2016-03-02
Use keypress event

bantic:
Use keypress event to accept character input

Previously, character input was detected after-the-fact using the mutation observer.

Other changes:

  * Add Logger class with `enable`/`disable`/`enableTypes` methods.
    Used to selectively turn on/off logging for different parts of the editor.
    Disabled by default
  * Add EditState class that is responsible for determining the current
    state of the editor (active markups, active sections, whether the next
    character will have any markup applied to it)
  * Refactor the timing of cursorDidChange callback events --
    * Do not fire `cursorDidChange` before the editor has rendered
    * Eagerly fire `cursorDidChange` when active markups or sections
      (but not necessarily the cursor) has changed. This provides compatibility
      for consumers that rely only on the `cursorDidChange` method to update the
      state of their toolbar buttons, e.g.
  * Change usage/lifecycle of `editor.range` -- now `editor.range` is cached until
    an event that may change the cursor (keystrokes, clicks, etc) occurs. This makes
    `editor.range` more stable and more performant. This obviates some complicated
    Adds `_resetRange` private method to force re-reading the range from DOM.
    `editor.range` is lazy — it only reads from the DOM when it is accessed.
  * Refactor `PostEditor` slightly to read and render editor's range in light of
    the changes to `editor.range`
  * Removes all usage of document.execCommand
  * Enables most tests that were turned off for IE. Only copy-paste tests are skipped
    in IE.
  * Adds test helper `skipInIE`
  * Adds `insertText` method to editor that inserts text at the current cursor position
    with the markups from that position (if the cursor is in bold text, more bold text will
    be inserted).
  * Adds EventManager class, moves event handling from Editor to EventManager
  * Adds `insertText` and `insertTextWithMarkup` to `PostEditor`
  * Change `Post#markupsInRange` to determine the markups correctly if the range is collapsed
  * Change `Key` to correctly determine if an event is a key event or is in printable in Firefox
    because Firefox reports `event.keyCode` as `0` for some key events.


bantic:
released in [v0.9.0](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.0)

PR #343, opened 2016-03-15
Ignore mutation events from within atoms

rlivsey:
ember-mobiledoc-editor uses ember-wormhole to insert components into 
atom placeholders. This immediately causes a mutation and then
the editor update is triggered when nothing has changed.

Atoms are their own world and so any changes in them should be safe
to ignore.

### DIFF
--- a/src/js/utils/dom-utils.js
+++ b/src/js/utils/dom-utils.js
@@ -92,6 +92,25 @@ function parseHTML(html) {
   return div;
 }
 
+function closest(el, selector, contentOnly=false) {
+  if (!isElementNode(el)) {
+    el = el.parentElement;
+  }
+  var matchesSelector = el.matches || el.webkitMatchesSelector || el.mozMatchesSelector || el.msMatchesSelector;
+
+  if (contentOnly && matchesSelector.call(el, selector)) {
+    return null;
+  }
+
+  while (el) {
+    if (matchesSelector.call(el, selector)) {
+      break;
+    }
+    el = el.parentElement;
+  }
+  return el;
+}
+
 export {
   containsNode,
   clearChildNodes,
@@ -103,5 +122,6 @@ export {
   isTextNode,
   isCommentNode,
   isElementNode,
-  parseHTML
+  parseHTML,
+  closest
 };
--- a/tests/unit/editor/atom-lifecycle-test.js
+++ b/tests/unit/editor/atom-lifecycle-test.js
@@ -234,3 +234,39 @@ test('onTeardown hook is called when editor is destroyed', (assert) => {
 
   assert.ok(teardown, 'onTeardown hook called');
 });
+
+test('mutating the content of an atom does not trigger an update', (assert) => {
+  const done = assert.async();
+
+  const atomName = 'test-atom';
+
+  const atom = {
+    name: atomName,
+    type: 'dom',
+    render() {
+      return makeEl('the-atom');
+    }
+  };
+
+  const mobiledoc = build(({markupSection, post, atom}) =>
+    post([markupSection('p', [atom(atomName, '@bob', {})])])
+  );
+  editor = new Editor({mobiledoc, atoms: [atom]});
+
+  let updateTriggered = false;
+  editor.on("update", function() {
+    updateTriggered = true;
+  });
+
+  assert.hasNoElement('#editor #the-atom', 'precond - atom not rendered');
+  editor.render(editorElement);
+
+  $("#the-atom").html("updated");
+
+  // ensure the mutations have had time to trigger
+  // TODO - nicer way of waiting?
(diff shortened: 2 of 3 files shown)


rlivsey:
If this is good, we might need to do the same for cards as they are inserted in the same way using wormholes in ember-mobiledoc-editor?

rlivsey:
Failure is saucelabs

bantic:
Thanks for this, makes total sense to me. I am wondering if it would be easier to delegate the responsibility to the render node the way we do for sections with `renderNode.reparsesMutationOfChildNode`.
I'll take a look at those sauce failures. I restarted the CI build for this yesterday but it seems to have failed in the same way again.

rlivsey:
Makes sense for this to be a renderNode concern, I'll take a look at that

src/js/utils/dom-utils.js
```diff
@@ -92,6 +92,25 @@ function parseHTML(html) {
   return div;
 }
 
+function closest(el, selector, contentOnly=false) {
+  if (!isElementNode(el)) {
+    el = el.parentElement;
```
bantic: This needs to be `el = el.parentNode` in order for IE to pass. IE doesn't define `parentElement` on `Node`, only on `Element` (and a textNode is a `Node` but not an `Element`). (Thanks a lot, IE.)

src/js/utils/dom-utils.js
```diff
@@ -92,6 +92,25 @@ function parseHTML(html) {
...
+  }
+  var matchesSelector = el.matches || el.webkitMatchesSelector || el.mozMatchesSelector || el.msMatchesSelector;
```
bantic: `let` :)

src/js/utils/dom-utils.js
```diff
@@ -92,6 +92,25 @@ function parseHTML(html) {
...
+  }
+
+  while (el) {
+    if (matchesSelector.call(el, selector)) {
+      break;
+    }
+    el = el.parentElement;
+  }
+  return el;
+}
```
bantic: It's probably very fast to traverse up the DOM tree like this, but it seems like it might be worthwhile to abort after we hit the editor element. An editor could be nested arbitrarily deeply in DOM. Passing in a final `stopElement` argument from the mutation handler would be nice.

tests/unit/editor/atom-lifecycle-test.js
```diff
@@ -234,3 +234,39 @@ test('onTeardown hook is called when editor is destroyed', (assert) => {
...
+  editor.render(editorElement);
+
+  $("#the-atom").html("updated");
+
+  // ensure the mutations have had time to trigger
+  // TODO - nicer way of waiting?
+  setTimeout(function(){
+    assert.ok(!updateTriggered);
+    done();
+  }, 100);
```
bantic: Curious if other smaller timeout values wouldn't be sufficient here.
Testing that the `didUpdatePost`, `didRender` and `willRender` hooks did not fire would be good coverage here, too.

bantic:
@rlivsey I added a few comments. This looks good to me. I will do a mini spike to try finding a render node (instead of a section render node) for the mutations to check the feasibility of simply delegating the `shouldReparse` logic to the render node itself, that feels a bit cleaner and also potentially more future-safe.

tests/unit/editor/atom-lifecycle-test.js
rlivsey: a time of 0 might be enough to run it in the next loop, I'll give it a go

src/js/utils/dom-utils.js
rlivsey: Yep that makes a lot of sense

src/js/utils/dom-utils.js
rlivsey: This definitely wasn't cut & paste from StackOverflow…

rlivsey:
@bantic thanks, I'll make those tweaks later tonight

bantic:
@rlivsey When you have a chance could you test this branch out? Moving reparsing logic onto the render nodes seems to work reasonably well:
https://github.com/bustlelabs/mobiledoc-kit/compare/master...ignore-atom-mutations-via-render-nodes

src/js/utils/dom-utils.js
bantic: Nice :smile:

bantic:
@rlivsey Ping. Did you have a chance to check out the spike on the branch I mentioned before? If that works I'd be happy to fold it in and finish up work here. Would love to get this merged.

bantic:
@rlivsey Thanks for this PR! I rebased and added a few additions (delegating reparsing to render nodes, mostly) in #362. thanks!

PR #347, opened 2016-03-22
Ensure using arrow keys when text is selected moves cursor correctly

bantic:
Fixes an issue where selecting text and then hitting left/right would
move the cursor one character past the left/right side of the selection
instead of putting the cursor *at* the left/right side of the selection.

  * Remove `Range#moveFocusedPosition`
  * Add `Range#move` and `Range#extend`, add tests for same
  * Refactor EventManager's `keydown` method to use `Range` `#move` and `#extend`


bantic:
released in [v0.9.0](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.0)

PR #350, opened 2016-03-24
[CLEANUP] DRY copy/paste and drop event handling. Add editor#serializeTo

bantic:
* Add `editor#serializeTo(format)`
  * Add `editor#serializePost(post, format)`
  * Add deprecate helper
  * Add `post#trimTo(range)` and deprecate `post#cloneRange`.
   `cloneRange` returns a mobiledoc, which is confusing.
  * Enable all copy-paste acceptance tests in IE. They all now use
    the copy/paste mocks from the test helpers.
  * Add an IE-compatibility test for `getContentFromPasteEvent` to
    ensure it will use `window.clipboardData`
  * Add an IE-compatibility test for `setClipboardData` to
    ensure it will use `window.clipboardData`
  * Remove now-unused `supportsStandardClipboardAPI` test helper

cc @YoranBrondsema this turns on all tests in all environments. Adds IE-compatibility tests for the two methods that may interact directly with `window.clipboardData`.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -37,6 +37,8 @@ import { MOBILEDOC_VERSION } from 'mobiledoc-kit/renderers/mobiledoc';
 import EditHistory from 'mobiledoc-kit/editor/edit-history';
 import EventManager from 'mobiledoc-kit/editor/event-manager';
 import EditState from 'mobiledoc-kit/editor/edit-state';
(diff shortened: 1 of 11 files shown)


bantic:
released in v0.9.1

YoranBrondsema:
@bantic Sorry for my late reply, was on a holiday last week. This looks great, thanks for enabling all tests!

PR #351, opened 2016-03-24
[BUGFIX] up/down arrows in Firefox should not update mobiledoc

bantic:
Firefox fires `keypress` events for up/down arrows (Chrome, Safari and IE11 do not), and `Key#isPrintable` was incorrectly returning true for these events because `String.fromCharCode(0).length > 0`. This updates `Key#isPrintable` to return false when the source event was an arrow keypress.


bantic:
released in v0.9.1

PR #353, opened 2016-03-24
[BUGFIX] Do not handle key events when the editor has no cursor

### LINKED ISSUE #344: Position#move isn't available on blank position
If the editor [handles a cursor key](https://github.com/bustlelabs/mobiledoc-kit/blob/c5292360ae2c8434abc4f47b293f4700891687ff/src/js/editor/editor.js#L643) when the editor doesn't have focus, then `editor.cursor.offsets` [returns a blank range](https://github.com/bustlelabs/mobiledoc-kit/blob/c5292360ae2c8434abc4f47b293f4700891687ff/src/js/utils/cursor.js#L62), with blank position objects. These blank positions don't have a `move` method and sadness occurs:

screenshot 2016-03-16 00 03 18

Should we just put a guard checking that `this.cursor.hasCursor()` at the top of [`handleKeydown`](https://github.com/bustlelabs/mobiledoc-kit/blob/c5292360ae2c8434abc4f47b293f4700891687ff/src/js/editor/editor.js#L624)?

bantic:
It is possible for the editor's element to be focused but not have a
cursor (selection). In this case, key events will fire on the editor's
element but the editor will not be able to determine a logical
`Position` (since there is no selection). When this happens, we abort
handling the key event.

fixes #344

cc @rlivsey

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -258,7 +258,7 @@ class Editor {
   }
 
   handleNewline(event) {
-    if (!this.cursor.hasCursor()) { return; }
+    if (!this.hasCursor()) { return; }
 
     event.preventDefault();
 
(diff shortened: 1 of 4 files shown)


bantic:
released in v0.9.1

PR #362, opened 2016-04-19
Ignore mutation events from within atoms

bantic:
ember-mobiledoc-editor uses ember-wormhole to insert components into
atom placeholders. This immediately causes a mutation and then
the editor update is triggered when nothing has changed.

Atoms are their own world and so any changes in them should be safe
to ignore.

Delegates the mutation-reparsing logic to RenderNodes.


bantic:
This adds the final touches to #343. Thanks @rlivsey !

bantic:
released in [v0.9.4](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.4)

PR #363, opened 2016-04-19
Allow destroying an editor when it failed to render properly

bantic:
If the editor throws an error while rendering an atom, later calling
`editor.destroy` can create issues. This ensures that the editor-dom
renderer doesn't choke when destroying an atom renderNode if it didn't
properly render.


bantic:
released in [v0.9.4](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.4)

PR #368, opened 2016-04-21
Add Editor#onTextInput hook, deprecate Editor#registerTextExpansion

### LINKED ISSUE #367: add editor hook to listen for entered text
In order to implement something like a system where typing "@" triggers an editor event that a consumer can listen to and then later replace with an atom (to represent that @-mention in some semantic/metadata-y sense).

The sort of flexible input configuration that [Mousetrap](https://craig.is/killing/mice) uses seems like a good path.

This system should handle:
  * inserted text (if the user types "abc", or "@" e.g.) — callback should be invoked after the text is inserted
  * keydown or keyup with and without modifiers ("command+k") — callback should be invoked on key down/up event and cancel default behavior

In the callback the following information should be available:
  * cursor position (probably just read this from `editor.range`)
  * have access to the `editor`
  * be able to specify that other listeners for the same text/key-combo *should* fire (by default only the first listener would likely fire)

bantic:
Adds `Editor#onTextInput`, which accepts an object with the following properties:
  * `text` — string; required if `match` is not present
  * `match` — regex; required if `text` is not present
  * `run` — callback

After a user types a character, the text preceding the cursor is compared to the input handlers. A handler is a match if the `text` prop is present and the preceding text ends with `text`, or if `match` is present and the preceding text matches `match`. The results of `match.exec(userEnteredText)` is passed to the handler's `run`.

This PR also deprecates `Editor#registerTextExpansion` because `onTextInput` covers the same use cases as text expansions and has a simpler API.

To add an @-mention handler one could do:
```
editor.onTextInput({
  text: '@',
  run(editor /*, matches=['@'] */) {
    let { range } = editor;
    let userName = window.prompt('Mention whom?');

    editor.selectRange(range.extend(-1)); // select the '@' character before the cursor
    editor.insertAtom('mention-atom', userName); // replaces the selected range with the atom
  }
});
```

cc @rlivsey

fixes #367


bantic:
This can also be used to auto-format text. For example, this makes text with "*" on either side of it *bold*, like markdown, and auto-linkifies URLs:

```
editor.onTextInput({
  match: /(https?:\/\/[^ ]+) $/,
  run(editor, matches) {
    let url = matches[1];
    let range = editor.range;
    editor.selectRange(range.move(-1).extend(-1 * url.length)); // select the URL
    editor.run(postEditor => {
      let link = postEditor.builder.createMarkup('a', {href: url});
      postEditor.toggleMarkup(link);
    });
    
    editor.selectRange(range); // restore original cursor position
  }
});

editor.onTextInput({
  match: /\*([\w\s]+)\* $/,
  run(editor, matches) {
    let fullText = matches[0];
    let text = matches[1];
    let range = editor.range;
  
    editor.run(postEditor => {
      let strong = postEditor.builder.createMarkup('strong');
      let position = postEditor.deleteRange(range);

      position = postEditor.insertTextWithMarkup(position, text, [strong]);
      postEditor.insertTextWithMarkup(position, ' ', []); // restore non-bold space after the text
    });
  }
});
```

demo

bantic:
released in [v0.9.5](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.5)

lessless:
👍

PR #375, opened 2016-04-26
[bugfix] Ensure cursor is in li after "* " expands to li

bantic:

PR #376, opened 2016-04-26
Ensure cursor positioning on a blank markup section or list item works

bantic:
`Cursor#_findNodeForPosition` should return the cursor element ("br",
usually) for a blank section. This change also ensures that the built-in
text-expansions for list items and headers position the cursor properly.

See also #375 which addressed this for list-items only.


bantic:
released in [v0.9.5](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.5)

PR #378, opened 2016-04-27
Consider HOME and END keys as movement keys

bantic:
fixes #377

PR #380, opened 2016-05-04
Added support to CMD+Left/Right for Mac navigation

atonse:
In MacOS, in addition to CTRL+A and E for "Home" and "End", another commonly used combination is CMD+Left for Home, and CMD+Right for End.

This commit adds those shortcuts, and pulls out the "getStartOfLine" and "getEndOfLine" functions out so they can respond to multiple command keys.

Regarding tests: I didn't find any tests for the existing Home/End key commands, so I didn't add anymore. But I have run the test suite and have noted no regressions.

### DIFF
--- a/tests/acceptance/editor-key-commands-test.js
+++ b/tests/acceptance/editor-key-commands-test.js
@@ -1,6 +1,7 @@
 import { Editor } from 'mobiledoc-kit';
 import { MODIFIERS } from 'mobiledoc-kit/utils/key';
 import Keycodes from 'mobiledoc-kit/utils/keycodes';
+import Browser from 'mobiledoc-kit/utils/browser';
 import Helpers from '../test-helpers';
 
 const { module, test } = Helpers;
@@ -63,7 +64,7 @@ function testStatefulCommand({modifierName, key, command, markupName}) {
     editor.render(editorElement);
 
     assert.hasNoElement(`#editor ${markupName}`, `precond - no ${markupName} text`);
-    Helpers.dom.moveCursorTo(editor, 
+    Helpers.dom.moveCursorTo(editor,
       editor.post.sections.head.markers.head.renderNode.element,
       initialText.length);
     Helpers.dom.triggerKeyCommand(editor, key, modifier);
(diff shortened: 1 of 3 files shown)


bantic:
@atonse Thanks for this PR! I *think* this was addressed by #378 . That code hasn't been released yet, though.

Cursor movement via key presses is one of the few places in mobiledoc-kit that we still allow some of the native browser behavior. This is mostly done because the heuristics involved to position correctly from up/down keys are challenging (where your cursor moves when you hit up-arrow is dependent on current cursor position, font metrics, and also where the cursor was previously on that line) so we let the native browser cursor movement happen and then [read the window's `Selection` from DOM again](https://github.com/bustlelabs/mobiledoc-kit/blob/019dd58ed1ee0a31c764e2fbefad6f36e56df2ea/src/js/editor/event-manager.js#L158-L160) if the key was a [movement key](https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/utils/key.js#L97-L99).

That said, explicitly handling home/end movement seems like a good idea, rather than trusting that the browser will put the cursor in the place we expect it to.

Adding key commands for cmd-up and cmd-down also seems good. These would focus on `post.headPosition()` and `post.tailPosition()`, respectively.

bantic:
Looking again now, I see that #378 deals with `fn+LEFT` (keyCode 36 aka `Home`) and `fn+RIGHT` (keyCode 35 aka `End`), so it is not the same.

bantic:
An acceptance test like [this one](https://github.com/bustlelabs/mobiledoc-kit/blob/019dd58ed1ee0a31c764e2fbefad6f36e56df2ea/tests/acceptance/editor-key-commands-test.js#L170) would be the ideal sort of test for these commands.

atonse:
@bantic the other navigational element is Alt/Option+left and right (go back and forward by a word). It was easy to add this PR since I just had to copy CTRL A/E behavior. But if you could point me to the right direction (like what you normally do to walk back until you find the next whitespace), I could add these other two.

Lastly, Shift + all of these should continue adding to the selection.

I don't envy you guys – this long tail of subtle behaviors can be endless.

bantic:
> confirm that CMD Up and Down at least moves to the top and bottom (including preserving the highlighted selection) but does not allow you to delete the selected text. I think in that sense, this is the same bug as #378.

Agreed. 😢  It seems that no `keyup` event fires when releasing the up/down arrow while <kbd>command</kbd> is also held down (on Chrome on my Mac), so the editor never re-reads the DOM position. Even more reason to deal with these explicitly/semantically.

> the other navigational element is Alt/Option+left and right (go back and forward by a word)

We have a [`section#textUntil(position)`](https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/models/_markerable.js#L44) method that would make going backwards simple, but there isn't (currently) a `textAfter` method.

If you could add an acceptance test for the <kbd>command</kb>-left and <kbd>command</kbd>-right I'd be happy to merge this PR as is and address the other movement combos separately. And I'd be very happy to help out on the follow-up here on github or through the [gitter chat](https://gitter.im/bustlelabs/mobiledoc-kit).

atonse:
Ok thanks @bantic, i'll work on the test and get back to you.

atonse:
Weird – I tested this locally using phantom (`npm test`) and all passed.

atonse:
From my local testem:
screen shot 2016-05-05 at 10 44 56 am

atonse:
Just realized this is MacOS only and the travis tests are failing for windows. I also added Firefox to my testem and have all local browsers passing. let's see how we do on windows.

atonse:
All out of ideas – any suggestions welcome :) 
screen shot 2016-05-05 at 11 07 56 am

tests/acceptance/editor-key-commands-test.js
```diff
@@ -140,6 +141,58 @@ testStatefulCommand({
...
+  let originalCursorPosition = Helpers.dom.getCursorPosition();
+  Helpers.dom.triggerKeyCommand(editor, 'RIGHT', MODIFIERS.META);
+
+  let changedCursorPosition = Helpers.dom.getCursorPosition();
+  let expectedCursorPosition = initialText.length; // end of text
+
+  if (Browser.isMac()) {
+    assert.equal(changedCursorPosition.offset, expectedCursorPosition, 'cursor moved to the end of the line on MacOS');
+  } else {
+    assert.equal(changedCursorPosition.offset, originalCursorPosition.offset, 'cursor not moved to the end of the line (non-MacOS)');
```
bantic: I think for windows this should actually assert that the cursor moves one character, not that it doesn't move at all.

bantic:
@atonse I'll give this a run-through in windows quickly and let you know what I see. :)

tests/acceptance/editor-key-commands-test.js
atonse: So I'm wondering about that because for me, Command on windows is the Windows Key, and Windows Key Left/Right actually are OS-level commands that rearrange your windows.

tests/acceptance/editor-key-commands-test.js
```diff
@@ -140,6 +141,58 @@ testStatefulCommand({
...
+  let textElement = editor.post.sections.head.markers.head.renderNode.element;
+
+  Helpers.dom.moveCursorTo(editor, textElement, 4);
+  let originalCursorPosition = Helpers.dom.getCursorPosition();
+  Helpers.dom.triggerKeyCommand(editor, 'LEFT', MODIFIERS.META);
+
+  let changedCursorPosition = Helpers.dom.getCursorPosition();
+  let expectedCursorPosition = 0; // beginning of text
+
+  if (Browser.isMac()) {
```
bantic: ah ha! `isMac` is not a function, it's a property. 😝

atonse: Oops! Sorry about that. fixed and pushed.

bantic:
@atonse looks great! Can you please squash the commits?

atonse:
@bantic squashed the commits.

bantic:
@atonse thank you!

bantic:
released in [v0.9.6](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.6)

PR #382, opened 2016-05-05
Add didrender hook

joshfrench:
Fires once card has been rendered, for use by cards that need processing after they've been added to the DOM (e.g. 3rd-party widgets.)

This is my first time in this codebase, so I don't expect this to be complete. Specifically:

- Did I overlook any cases where this hook should (or should not) fire?
- Would it be useful for the callback take any args, such as the current `env`?
- I based this on the `onTeardown` hook. Is the `env.didRender` pattern preferable to adding another optional `didRender` property to the top-level card object?

Thanks for your feedback!

### DIFF
--- a/src/js/models/card-node.js
+++ b/src/js/models/card-node.js
@@ -46,11 +46,18 @@ export default class CardNode {
     }
   }
 
+  didRender() {
+    if (this._didRenderCallback) {
+      this._didRenderCallback();
+    }
+  }
+
   get env() {
     return {
       name: this.card.name,
       isInEditor: true,
       onTeardown: (callback) => this._teardownCallback = callback,
+      didRender: (callback) => this._didRenderCallback = callback,
       edit: () => this.edit(),
       save: (payload, transition=true) => {
         this.section.payload = payload;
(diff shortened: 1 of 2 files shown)


bantic:
This looks great! I restarted the travis build.

bantic:
@joshfrench This looks great, thank you for the PR.
I'm going to merge and then update the docs to mention this hook.
Because it's possible for folks to reuse cards between editor and display-only rendering contexts, I want to be clear that the hook is only available for cards that are rendered by the editor (and not by, e.g., [mobiledoc-dom-renderer](https://github.com/bustlelabs/mobiledoc-dom-renderer)). I'm not sure it makes as much sense for that renderer, anyway, because the mobiledoc-dom-renderer builds the entire dom fragment for the mobiledoc before appending anything.

joshfrench:
Agreed, I wouldn't expect the downstream renderer to have to handle this. As for callback args, am I right in thinking `env` and friends are unnecessary since they're available via closure when the callback is defined?

bantic:
> am I right in thinking env and friends are unnecessary since they're available via closure when the callback is 

Yeah, I think's exactly right. All possible variables they might need are in scope already, so they can close over them in their callback function. 👍

bantic:
released in [v0.9.6](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.6)

PR #394, opened 2016-05-11
Fix bug #395: Error when adding a selection outside the editor while not editing

### LINKED ISSUE #395: Error when adding a selection outside the editor while not editing
When the editing mode is disabled, adding a selection outside of the editor (for example by triple clicking) throws an error.

````
Error: Could not find parent section for un-mapped text node "outside section"
````

I started a PR with a failing test: https://github.com/bustlelabs/mobiledoc-kit/pull/394.

However, before I continue, @bantic maybe you can tell me what the best way to proceed would be. My thinking would be to exclude any selection outside of the editor. So when a selection includes a node after the editor, limit the selection to the last node in the editor. And if a selection includes a node from before the editor, limit the selection to the first node in the editor.

This will however, prevent from selecting anything outside of the editor. Alternatively, one could "split" the selection in two. Have the editor part handled by mobiledoc-kit and leave the other part as it is. I'm not sure how feasible that is so I'd like to have your input.

### PR DESCRIPTION
**Pull request by YoranBrondsema**
Fixes #395

### DIFF
--- a/tests/acceptance/editor-selections-test.js
+++ b/tests/acceptance/editor-selections-test.js
@@ -71,8 +71,10 @@ test('when editing is disabled, the selection detection code is disabled', (asse
                          'outside', outsideSection);
 
   Helpers.dom.triggerEvent(document, 'mouseup');
+
   assert.equal(editor.activeSections.length, 0, 'no selection inside the editor');
-  assert.equal(Helpers.dom.getSelectedText(), "trick pony\n\noutside");
+  // Edge and IE11 do not introduce newlines
+  assert.equal(Helpers.dom.getSelectedText().replace('\n', ''), "trick ponyoutside");
 });
 
 test('selecting an entire section and deleting removes it', (assert) => {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/acceptance/editor-selections-test.js [hunk 1]
```diff
@@ -57,6 +57,26 @@ test('selecting across sections is possible', (assert) => {
...
+  let outsideSection = $('p:contains(outside section)')[0];
+
+  Helpers.dom.selectText(editor ,'trick', firstSection,
+                         'outside', outsideSection);
+
+  Helpers.dom.triggerEvent(document, 'mouseup');
+
+  assert.equal(editor.activeSections.length, 0, 'no selection inside the editor');
+  // Edge and IE11 do not introduce newlines
+  assert.equal(Helpers.dom.getSelectedText().replace('\n', ''), "trick ponyoutside");
```
bantic: This could be `assert.ok(...selectedText.indexOf('trick pony') !== -1 && selectedText.indexOf('outside') !== -1)` to assert that the selection includes inside and outside. Trying to wrangle the selection-to-text differences between browsers is like whack-a-mole. :)

YoranBrondsema: Makes sense.

YoranBrondsema:
@bantic Let me know if it's all good for you.

bantic:
👍 thank you!

bantic:
released in [v0.9.7](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.7)

PR #396, opened 2016-05-11
Handling emoji's on iOS Safari

YoranBrondsema:
iPads have emoji's on their virtual keyboard. However, this causes problems with mobiledoc-kit as both the `charCode` and the `keyCode` are 0 for the `keydown` and `keyup` events when inputting one of those emojis. To illustrate, I used the Keyboard Event Viewer (https://w3c.github.io/uievents/tools/key-event-viewer.html):

emoji

This causes the assert at https://github.com/bustlelabs/mobiledoc-kit/blob/3cb78311cc0ffc57067537488586a986e2179adc/src/js/utils/key.js#L75 to fail.

I solved it by changing the heuristic to determine whether an event is a key event to checking whether the type is one of `keypress`, `keydown` or `keybottom`. According to the [doc](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent), any key event has one of these values. Also, this is the heuristic that is used by jQuery (see https://github.com/jquery/jquery/blob/master/src/event.js#L627).

### DIFF
--- a/src/js/utils/key.js
+++ b/src/js/utils/key.js
@@ -51,11 +51,7 @@ export const SPECIAL_KEYS = {
 
 // heuristic for determining if `event` is a key event
 function isKeyEvent(event) {
-  return !!event.charCode ||
-    !!event.keyCode ||
-    !!event.metaKey ||
-    !!event.shiftKey ||
-    !!event.ctrlKey;
+  return /^key/.test(event.type);
 }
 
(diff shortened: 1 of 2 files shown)


bantic:
thank you, great work! Thanks for the thoroughness, and for showing that keyboard event viewer. I hadn't seen it before.

PR #414, opened 2016-06-06
Allow `target` attr on anchor tags

joshfrench:


joshfrench:
Oops, just found https://github.com/bustlelabs/mobiledoc-kit/issues/346. LMK if this patch isn't the direction you wanted to go in. Thanks!

mixonic:
@joshfrench we can revisit. I'm still a bit hesitant to embrace this. I am, however, totally supportive of a `markupElementRenderer` option for the DOM renderer akin to [sectionElementRenderer](https://github.com/bustlelabs/mobiledoc-dom-renderer#sectionelementrenderer). This would make it easy to add `target=` at render-time (which seems like the more common use-case anyway, render-time instead of edit-time).

joshfrench:
Reading through that thread, `markupElementRenderer` also sounds like the better alternative to me. I'll close this and take discussion over there.

PR #420, opened 2016-06-11
Render prettier mobiledocs in the demo app

courajs:
This makes the docs in the demo app easier to read:
screen shot 2016-06-11 at 1 33 39 pm
Compared to existing:
screen shot 2016-06-11 at 1 34 02 pm

The main drawback right now is that it was only really written for 0.3.0 docs. 0.2.0 docs look a little wonky:
screen shot 2016-06-11 at 1 38 53 pm

### DIFF
--- a/demo/app/helpers/format-object.js
+++ b/demo/app/helpers/format-object.js
@@ -1,7 +1,8 @@
 import Ember from 'ember';
+import formatMobiledoc from 'npm:mobiledoc-pretty-json-renderer';
 
 export function formatObject([object]) {
-  return JSON.stringify(object, null, '  ');
+  return formatMobiledoc(object);
 }
 
 export default Ember.Helper.helper(formatObject);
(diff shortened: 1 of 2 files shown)


mixonic:
@courajs niiiice! :-D Can I ask you to just change which way to render the output based on the version of the document? The document should have a version string you can check.

Some like: https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/has-ember-version.js to check the number.

courajs:
Sure, I'll fall back to `JSON.stringify(obj, null, 2)` for anything not `0.3.0`

mixonic:
👍 :-D

courajs:
`mobiledoc-pretty-json-renderer` now falls back to old behavior for non `0.3` documents

mixonic:
🙇 amazing. Thanks @courajs!

PR #423, opened 2016-06-22
It's useful to expose replace section functions

eguitarz:
In case of extending text input handlers by users, it would be useful not copy/pasting`replaceWithListSection`/`replaceWithHeaderSection`. For example if I just want to add a new handler to convert section to `backquote`, I can easily import replaceWithHeaderSection from this file without implementing this function again. So I propose to expose these two functions.


bantic:
@eguitarz Thanks for this PR. Apologies that it has lingered. I'm going to follow-up and document these functions as public.

PR #427, opened 2016-06-29
Update docs: didRender is defined on the card card, regardless of editing context

joshfrench:
See bustlelabs/mobiledoc-dom-renderer#31


bantic:
thanks @joshfrench !

PR #428, opened 2016-07-01
Return cards and atoms from high-level insert methods

joshfrench:
It can be useful to get a handle on newly inserted cards/atoms. Currently the only way to do that is to drop to the lower-level `postEditor` methods. This keeps the convenience of high-level insert behavior (range detection, nested/blank sections, &c) while enabling additional operations on the inserted element.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -908,6 +908,7 @@ class Editor {
    * @param {String} atomName
    * @param {String} [atomText='']
    * @param {Object} [atomPayload={}]
+   * @return {Atom} The inserted atom.
    * @public
    */
   insertAtom(atomName, atomText='', atomPayload={}) {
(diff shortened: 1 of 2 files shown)


mixonic:
Seems reasonable, FF failed on Sauce and I've kicked it

bantic:
Sauce still finicky, I kicked it again

bantic:
investigating why tests in FF @ Sauce failed. Runs fine locally.

bantic:
There's an issue with FF at Sauce where it doesn't properly focus the editor element in some situations. I can't reproduce it locally, but adding a line to force a selection in the `insertAtom` test fixes this problem at Sauce. I'll add that manually after a merge.

thanks @joshfrench !

bantic:
released in [v0.9.8](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.9.8)

PR #438, opened 2016-07-21
Add forward/backward inclusivity rules for markups fix #402 fix #392

### LINKED ISSUE #392: After pasting URL, hitting spacebar should not edit URL
Immediately after pasting a URL into editor, users should be able to hit the spacebar without adding to the URL. 

test

### LINKED ISSUE #402: Prevent including typed text into the existing markup
Hello, 

I'm really struggling to find proper words (for the title as well), so let me quote @bantic's answer on SO
> default behavior for browsers (and for mobiledoc-kit, so far) is for the style of text to be "right-inclusive". If the cursor is on the right side of styled text, typing continues that style. 

While this might good for the text formatting commands, it's really making user experience more complicated in some situations like link insertion.

Let's take demo app as example:

1. there is a text 'This is Mobiledoc-kit' on the page
2. I select 'Mobiledoc-kit' fragment of it and make it a link pointing to the https://bustlelabs.github.io/mobiledoc-kit/demo/
3. Right after that I started to type and entered ' example app'
4. And this text became a part of the link, so now link is not just 'Mobiledoc-kit', but also all the text till the end of the line: 'Mobiledoc-kit example app'

I know that there is no ready answer yet, so I just want to put that issue on the order of business.

Also,  might it be sensible to make this behavior adjustable, not just for links but maybe for all commands?

bantic:
Changes `Post#markupsInRange` to check the markers to the left and right
of the position when range is collapsed. The markups in those markers
provide `isForwardInclusive` and `isBackwardInclusive` methods that are
used to determine which markups should be considered "active" at that
range.

fix #392
fix #402

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -475,8 +475,11 @@ class Editor {
   /**
    * @param {Markup|String} markup A markup instance, or a string (e.g. "b")
    * @return {boolean}
+   * @deprecated after v0.10.1
    */
   hasActiveMarkup(markup) {
+    deprecate(`editor#hasActiveMarkup is deprecated. Use editor#activeMarkups instead`);
+
     let matchesFn;
     if (typeof markup === 'string') {
       markup = markup.toLowerCase();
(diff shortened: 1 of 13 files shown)


lessless:
thank you! :heart_eyes:

PR #448, opened 2016-08-04
Use "version" script instead of postversion with --amend --no-edit

bantic:
Using --amend --no-edit was changing the sha of the commit *after* it
had been tagged, which caused issues with future updates to the
changelog. The `git-semver-commits` package failed to find tags when
their commits were amended post-tagging.


bantic:
released in [v0.10.4](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.4)

PR #449, opened 2016-08-05
Add nobreak to links in the editor so toolips are always accessible

gdub22:
There is case where if a long link wraps, the tooltip to view the outbound link may not be accessible via the mouse.

Before:
screen shot 2016-08-05 at 3 31 54 pm

After:
screen shot 2016-08-05 at 3 50 32 pm


bantic:
+1 good catch, @gdub22

bantic:
released in https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.5

Issue #171, opened 2015-10-13
Placeholder doesn't appear when content is deleted


rlivsey:
[Post#isBlank](https://github.com/bustlelabs/content-kit-editor/blob/master/src/js/models/post.js#L17) returns false if you type something and then delete all the content resulting in the placeholder text not re-appearing.

The current implementation checks that there are no sections, but I'm guessing at this point there's one blank section.

Should this change to be blank if there are no sections OR one blank section?

Post#isBlank looks to be [used internally](https://github.com/bustlelabs/content-kit-editor/blob/f3feeee09d3e10974b8af2ce48c874283cf752a1/src/js/editor/editor.js#L583) in a few places, so not sure if that would break those.

Alternatively, there could be a public Editor#isBlank which is to be used instead & change the placeholder behaviour to use this?

rlivsey:
Example mobiledoc from the demo before/after deleting all content:

```json
```

```json
{
  "version": "0.2.0",
  "sections": [
    [],
    [
      [
        1,
        "p",
        []
      ]
    ]
  ]
}
```

bantic:
@rlivsey Good question. Some of the code surrounding the `post#isBlank` was to work around the difficulty of showing a placeholder when also showing a "cursorable" `<br>` (in order to make it so the cursor can be shown reliably, an empty markup section is rendered as a `<p><br></p>`). 

It may make the most sense to preserve the implementation of `post#isBlank` and instead ensure that when removing the last text from the only markup section it is also removed. That also allows the placeholder to reappear, which seems good. It also preserves a 1-1 mapping of a mobiledoc with a blank post, rather than having two visually-indistinguishable "blank" mobiledocs (your before/after mobiledocs, above).

What do you think?

rlivsey:
@bantic yep that makes complete sense, I'll see if I can figure out how to make it do that…

mixonic:
it is probably a little wonky, and will take some work on the renderer. I'd be happy to take a look next week.

bantic:
closed by #461

PR #461, opened 2016-08-23
Use '__has-no-content' class name to display placeholder text

### LINKED ISSUE #407: Placeholder doesn't  re-appear
Hello,

Everybody loves  placeholders! They help to recognize typing area and get work started.

Unfortunately, Mobiledoc-kit's placeholder is rather shy or lazy: after it showed up he will never come back, or as @mixonic said:
>  they don't detect empty doc

How to reproduce:

1. Go to https://bustlelabs.github.io/mobiledoc-kit/demo/
2. Change example to "Empty mobiledoc"
3. Observe "Write here" placeholder
4. Type something
5. Erase all the text

Expected behavior: placeholder should re-appear
Actual behavior: placeholder doesn't re-appear 

Let's make placeholder great again! :v:

**UPDATE:** actually, on my machine placeholder doesn't appear even in this minimalist setup https://github.com/lessless/mobiledoc-placeholder

bantic:
This causes the placeholder to appear even when the editor div is not
wholly devoid of elements.

Possible BREAKING CHANGE: The placeholder text is now absolutely
positioned (at `top: 0`) of the editor div. If a user has applied their
own padding to the editor div the placeholder may not show up in the
correct location. The solution is to add a CSS rule that overrides the
`top` to match the value of the `padding-top`:
```
.__mobiledoc-editor.__has-no-content:after {
  top: <value of padding-top>;
}
```

Also: The editor now has a `min-height: 1em;`
to ensure that the placeholder has a space in which to be displayed.

Fixes #407 #171

### DIFF
--- a/src/css/editor.less
+++ b/src/css/editor.less
@@ -10,24 +10,23 @@
   font-family: 'Lora', Georgia, serif;
   margin: 1em 0;
   color: #454545;
-  /*
-    Chrome bug adds inline styles when backspacing to join 2 blocks.
-    Fix: Apply font styles to parent element, or use % for font-size, line-height.
-    http://stackoverflow.com/questions/15015019/prevent-chrome-from-wrapping-contents-of-joined-p-with-a-span
-  */
-  font-size: 120%;
-  line-height: 160%;
+  font-size: 1.2em;
+  line-height: 1.6em;
+  position: relative;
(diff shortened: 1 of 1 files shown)


bantic:
released in `v0.10.8-beta.1`

PR #466, opened 2016-08-24
Fix insertion into blank section

### LINKED ISSUE #462: Pasting text that contains a newline throws an error
I think this is a regression introduced by #459. Inserting text with newlines inside a list works, as per the fix. However, inserting text with newlines outside of a list, for instance in an empty Mobiledoc instance, fails.

error

bantic:
* Refactor #insertPost tests to use builder and editor DSL.
* Add test helper editor DSL `Helpers.editor.buildFromText(dsl, editorOptions)`
* Add test helper `retargetRange` to retarget a range from one post to
  another. Helps with asserting that the editor's range matches the one
  from the DSL

Fixes #462

### DIFF
--- a/src/js/editor/post/post-inserter.js
+++ b/src/js/editor/post/post-inserter.js
@@ -66,11 +66,7 @@ class Visitor {
     if (this.cursorSection.isBlank && !this._isNested) {
       // replace blank section with entire post
       let newSections = node.sections.map(s => s.clone());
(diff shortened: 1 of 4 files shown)


bantic:
released in [v0.10.7](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.7)

PR #471, opened 2016-08-25
Avoid rerendering atoms when section content changes.

### LINKED ISSUE #421: Atoms re-render whenever parent node updates (?)
Once I've added an atom, its `render()` method is fired on any subsequent changes in the same section. For the Ember editor (and now the React version :grin:) this means the `ADD_ATOM_HOOK` also fires again, and you end up with N+1 objects in `componentAtoms` which are never cleaned up.

Not sure if this is by accident or design, since cards don't exhibit this behavior (or more accurately can't, because you can't edit their node.) If it is intentional I'll open a bug in the Ember editor for cleaning up the orphaned `componentAtoms`.

atom

bantic:
fixes #421

### DIFF
--- a/src/js/models/atom-node.js
+++ b/src/js/models/atom-node.js
@@ -13,16 +13,13 @@ export default class AtomNode {
   }
 
   render() {
-    this.teardown();
+    if (!this._rendered) {
+      let {atomOptions: options, env, model: { value, payload } } = this;
+      // cache initial render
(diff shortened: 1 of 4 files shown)


bantic:
released in [v0.10.8](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.8)

PR #476, opened 2016-08-30
Allow reading range from DOM when editor is disabled

### LINKED ISSUE #475: moving cards in disabled editor can cause `editor.range` to go out of sync
In a disabled editor, the `_readRangeFromDOM` method no-ops. This can cause `editor.range` to stay cached on an old state. If that range references sections that have since been programmatically deleted from the post, unexpected errors can happen, for example when snapshotting the range (to add to the undo stack), the `range.head.section` is expected to have a `post` property, but if that section has been removed it no longer will. This causes an error in [the `position#leafSectionIndex` code](https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/utils/cursor/position.js#L120).

bantic:
Because `_readRangeFromDOM` doesn't do anything
if the editor is non-editable, the editor may get stuck with a range
that is pointing at a section that has been removed from the post.
This causes the error seen when doing the `post.walkAllLeafSections`
later in the snapshot code. The post is undefined because it is read off
the card, but the card that the range points at has been removed.

Fixes #475

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -395,9 +395,6 @@ class Editor {
   }
 
   _readRangeFromDOM() {
-    if (!this.isEditable) {
-      return;
-    }
     this.range = this.cursor.offsets;
   }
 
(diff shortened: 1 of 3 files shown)


bantic:
released in [v0.10.9](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.9)

PR #479, opened 2016-08-30
Ensure `onTextInput` is triggered by tab character

### LINKED ISSUE #400: onTextInput not triggered on tab
quoting @bantic 

> This can probably be fixed simply by removing the special-handling of tab in the keydown handler here https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/editor/event-manager.js#L142

bantic:
Change the event manager to delegate insertion of tab key to the input
handler rather than simply inserting the string.

Also: Lot of cleanup to tests.

Fixes #400


bantic:
released in [v0.10.9](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.9)

PR #484, opened 2016-09-02
Able to unregiser single or all text input handlers

### LINKED ISSUE #480: Unable to disable default onTextInput handlers
Default textInputHandlers are nice but I found there's no public way to disable or override them.

For now I'm doing this to disable default handlers:
```
editor._eventManager._textInputHandler.destroy();
```

But there should be a public method in editor to do this. Maybe `resetTextInputHandlers()`?

eguitarz:
This PR resolves #480

### DIFF
--- a/src/js/editor/event-manager.js
+++ b/src/js/editor/event-manager.js
@@ -45,6 +45,15 @@ export default class EventManager {
     this._textInputHandler.register(inputHandler);
   }
 
+  unregisterInputHandler(name) {
+    this._textInputHandler.unregister(name);
+  }
+
+  unregisterAllTextInputHandlers() {
+    this._textInputHandler.destroy();
+    this._textInputHandler = new TextInputHandler(this.editor);
+  }
+
   _addListener(context, type) {
     assert(`Missing listener for ${type}`, !!this[type]);
 
--- a/README.md
+++ b/README.md
@@ -241,6 +241,11 @@ editor.onTextInput({
 The editor has several default text input handlers that are defined in
 `src/js/editor/text-input-handlers.js`.
 
+To remove default text input handlers, simply call the unregister function.
+```javascript
+editor.unregisterAllTextInputHandlers();
+```
+
 ### DOM Parsing hooks
 
 A developer can override the default parsing behavior for leaf DOM nodes in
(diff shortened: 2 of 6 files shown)


bantic:
@eguitarz Thanks, this looks great! Only change is can you change `_validateHandler` to allow a handler without a name but add a deprecate to it, e.g. `deprecate('Registered input handlers require a "name' property so that they can be unregistered', !!handler.name)`? thanks!

eguitarz:
Sounds good! It's a bit too aggressive change the interface without notice. I'll make an update tonight, thanks!

eguitarz:
@bantic Please 👀

src/js/editor/event-manager.js
```diff
@@ -45,6 +45,14 @@ export default class EventManager {
     this._textInputHandler.register(inputHandler);
   }
 
+  unregisterInputHandler(name) {
+    this._textInputHandler.unregister(name);
+  }
+
+  unregisterAllTextInputHandlers() {
+    this._textInputHandler.destroy();
```
bantic: This needs to create a new text input handler afterward, right? So that someone can register new handlers after the unregister, like this:
```
editor.unregisterAllTextInputHandlers();
editor.onTextInput({...}); // there needs to be a non-destroyed input handler
```

bantic:
@eguitarz Looks good, just one note about re-creating the input handler after destroying it. Sorry for the delay. If you can make that change I think this is good to go. Thanks!

src/js/editor/event-manager.js
eguitarz: yes, `destroy()` in textInputHandler actually is not destroying the `textInputHandler`, it's assigning an empty array to textInputHandler. 

But for clearer purpose, I will assign a new textInputHandler after destroying.

eguitarz:
@bantic Addressed the review comments.

src/js/editor/event-manager.js
bantic: Ah, yes, you're right. Although in the future I could see the implementation of `destroy` changing in a way that makes it not possible to continue to register new handlers, so re-creating it feels good. thanks!

bantic:
released in [v0.10.10](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.10). thanks @eguitarz !

Issue #486, opened 2016-09-05
Triple-clicking to select throws an error when editing is disabled


YoranBrondsema:
The problem is that when selecting by triple-clicking, the tail of the selection is outside of the editor. I added a failing test in [this PR](https://github.com/bustlelabs/mobiledoc-kit/pull/487/files).

![bug](https://cloud.githubusercontent.com/assets/1374412/18253600/a1e6299c-735b-11e6-8c69-920498ff2258.png)

I think it can be solved by identifying this case and then limiting the selection to the last possible position inside the editor (I guess the direction of the selection has to be taken into account as well).

bantic:
@YoranBrondsema thanks as always for the very thorough bug report (and failing test). I'll take a look at this. Should be possible to tighten up `Position.fromNode` to better deal with this case.

PR #487, opened 2016-09-05
Fix bug #486

YoranBrondsema:


bantic:
Thanks for putting this example test together. #542 fixes the bug shown here (I opted for a more fine-grained test that used the native selection API to select the exact nodes that Chrome and Safari would select on a triple-click).

Issue #481, opened 2016-08-31
feature: Able to match newline in onTextInput handler


eguitarz:
When implementing heading (H1, H2, H3) in WYSIWYG markdown editor, it would requires detecting if `newline` is entered. So I can do something like this:
```
editor.onTextInput({
  // "# " -> h1, "## " -> h2, "### " -> h3
  match: /^(#{1,3})(.*)\n/,
  run() {
    // magic happens here
  }
});
```
Is this possible in current structure?

bantic:
@eguitarz Is this because you want to do the conversion only after hitting enter on the line? Would it work for your purposes to convert to h* after the user hits the space? e.g.:
  * user types "##"
  * user types " " -> "##" is deleted and section is converted to an empty h2 section with the cursor at the start of it

To answer you question, the "\n" is not matched by the `onTextInput` — <kbd>enter</kbd> is handled by `editor.insertNewline()` instead. We can certainly add hooks similar to the delete hooks for before/after inserting a newline. As discussed here (https://github.com/bustlelabs/mobiledoc-kit/issues/451#issuecomment-239449216), it would be useful for some folks to be able to have a little more control over what the `insertNewline` functionality means (the ability to keep or change the current  markup section tagname, e.g. — right now any newline resets to a standard "p" markup section).

bantic:
There's also a request for the ability to special-case newline-handling for different section  ypes here: https://github.com/bustlelabs/mobiledoc-kit/issues/432#issuecomment-239194830

eguitarz:
@bantic I agree this could be a temporary solution, but if we can control `insertNewline` would be great. Because when implementing image or link markdown syntax, it would be nice to insert a card or atom after you press `enter`. 
```
 ![alt text](https://github.com/adam-p/markdown-here/raw/master/src/common/images/icon48.png)
```

I believe this could make mobiledoc allows more interesting customizations.

bantic:
+1 yes, if you are interested in making a PR that'd be great. Otherwise I'll  mark this as a feature request and we'll get it in as soon as we can. 

--
twitter.com/bantic

eguitarz:
Thanks, will see what I can do to make this happen!

eguitarz:
@bantic After thinking about this feature, I'm proposing two modifications.
1. Add `willHandleNewline`/`didHandleNewline` hooks.
2. Add `disableNewline` function in editor

[...]

PR #489, opened 2016-09-07
add willHandleNewline hook

eguitarz:
- It addresses #481 
- I was planning to handle newline base on callback return value. But current callback structure omits the return values. Instead, I passed a parameter into `willHandleNewline` and handle newline base on its value is changed or not.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -63,6 +63,7 @@ const CALLBACK_QUEUES = {
   DID_RENDER: 'didRender',
   WILL_DELETE: 'willDelete',
   DID_DELETE: 'didDelete',
+  WILL_HANDLE_NEWLINE: 'willHandleNewline',
   CURSOR_DID_CHANGE: 'cursorDidChange',
   DID_REPARSE: 'didReparse',
   POST_DID_CHANGE: 'postDidChange',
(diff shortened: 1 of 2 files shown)


src/js/editor/editor.js
```diff
@@ -339,6 +340,12 @@ class Editor {
           return;
         }
       }
+
+      // Above logic might delete redundant range, so callback must run after it.
+      let options = { insertNewline: true };
```
bantic: @eguitarz Instead of using this ad hoc options hash, let's use a synthetic `event` object that has a `preventDefault` method on it. In user code, the developer can call the `preventDefault` method on it, e.g.:
```javascript
editor.willHandleNewline((event) => {
  if (shouldPreventDefault) {
    event.preventDefault();
  }
});
```

eguitarz:
@bantic Please check if this version is what you preferred. Unfortunately we have to do some workarounds to support custom events in IE, the code is a bit longer than I thought.

bantic:
@eguitarz Thanks for doing all this work! I think I inadvertently led you astray, I apologize. I just meant passing a basic event-like object with a `preventDefault` method on it:
```javascript
let defaultPrevented = false;
const event = { preventDefault() { defaultPrevented = true; };
this.runCallbacks(CALLBACK_QUEUES.WILL_HANDLE_NEWLINE, [event]);
if (defaultPrevented) { ...
```

I'd prefer that to using a true `CustomEvent` because of the inconsistencies between browsers.
If you can make that one change this will be good to go and I'd be happy to merge. Sorry about the back-and-forth.

eguitarz:
@bantic No problem. I enjoyed in this back-and-forth convo, so don't worry about it. Now, it should be the version you preferred.

bantic:
@eguitarz 👍 thank you for the hard work!

bantic:
released in [v0.10.10](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.10).

Issue #485, opened 2016-09-03
Export default `toggleLink` behavior?


joshfrench:
What do you think about exposing the default `toggleLink` handler for use in editor toolkits? I'm prepping the React editor for release and realizing I'd rather just use the core `META+K` dialog on the `Link` button instead of supplying our own form there.

Happy to submit a pull if you're up for it, although I'm not sure where the right place to export it would be. Thoughts on where best to expose it?

bantic:
@joshfrench Yeah, that certainly seems fine. +1 to code reuse.
Exporting a top-level "UI" might make sense for this. It would be good use this opportunity to change `toggleLink` to accept a configurable method for showing the prompt, i.e. pass it a callback-accepting function to use to show the prompt:
```
function toggleLink(editor, showPrompt=defaultShowPrompt) {
  ...
  let myPromptCallback = (url) => /* ... toggleMarkupWithUrl... */;
  showPrompt(message, defaultValue, myPromptCallback);
}

let defaultShowPrompt = (defaultValue, callback) => callback(window.prompt(message, defaultValue));
```

Exporting it as `Mobiledoc.UI.toggleLink` seems ok — what do you think?
If we do this we can also remove `Editor#showPrompt` as it is not used anywhere else.

joshfrench:
Looks good to me, I'll put together a patch this week. Thanks!

bantic:
closed by #491. thanks @joshfrench !

PR #491, opened 2016-09-09
Export toggleLink

joshfrench:
[re: #485] Makes `Mobiledoc.UI.toggleLink` a public function so that alternate editor kits can reuse and/or customize the core URL prompt. (I18n, anyone?)

Incidentally extracts `renderIntoAndFocusTail` into `test-helpers` as it was duplicated across several files.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -344,10 +344,6 @@ class Editor {
     });
   }
 
-  showPrompt(message, defaultValue, callback) {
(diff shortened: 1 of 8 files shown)


bantic:
Thanks @joshfrench this is excellent. Thanks for adding the thorough jsdoc, as well!

bantic:
released in [v0.10.10](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.10). thanks @joshfrench  !

PR #496, opened 2016-09-15
Add range to willHandleNewline hook

eguitarz:
I find out I have to access current section context to determine to disable newline or not. Seems add the editor and range into current hook callback is the easiest way.

The reason we also need editor, is to allow user to modify post in willHandleNewline callback.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -344,7 +344,7 @@ class Editor {
       // Above logic might delete redundant range, so callback must run after it.
       let defaultPrevented = false;
       const event = { preventDefault() { defaultPrevented = true; } };
-      this.runCallbacks(CALLBACK_QUEUES.WILL_HANDLE_NEWLINE, [event]);
+      this.runCallbacks(CALLBACK_QUEUES.WILL_HANDLE_NEWLINE, [event, this, range]);
       if (defaultPrevented) { return; }
 
       cursorSection = postEditor.splitSection(range.head)[1];
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -344,7 +344,7 @@ class Editor {
       // Above logic might delete redundant range, so callback must run after it.
       let defaultPrevented = false;
       const event = { preventDefault() { defaultPrevented = true; } };
-      this.runCallbacks(CALLBACK_QUEUES.WILL_HANDLE_NEWLINE, [event, this, range]);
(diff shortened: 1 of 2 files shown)


bantic:
@eguitarz Sorry for the delay here. Is there a reason you can't use the existing `editor` and read its `range` in the callback?
e.g.:

```javascript
editor.willHandleNewline(event => {
  let range = editor.range;
});
```

eguitarz:
@bantic I forgot why I added range, seems we can access editor.range directly... I'm going to change the commit a bit.

eguitarz:
@bantic Ready for you

src/js/editor/editor.js
```diff
@@ -344,7 +344,7 @@ class Editor {
       // Above logic might delete redundant range, so callback must run after it.
       let defaultPrevented = false;
       const event = { preventDefault() { defaultPrevented = true; } };
-      this.runCallbacks(CALLBACK_QUEUES.WILL_HANDLE_NEWLINE, [event]);
+      this.runCallbacks(CALLBACK_QUEUES.WILL_HANDLE_NEWLINE, [event, this]);
```
bantic: I still don't think we need to pass the editor again. It will be in scope in the user's code already.

tests/acceptance/basic-editor-test.js
```diff
@@ -214,7 +214,8 @@ test('keypress events when the editor does not have selection are ignored', (ass
 test('prevent handling newline', (assert) => {
   editor = Helpers.editor.buildFromText('', {element: editorElement});
 
-  editor.willHandleNewline(event => {
+  editor.willHandleNewline((event, {range}) => {
+    assert.equal(range.head.section.text, 'Line1', 'can access section text');
```
bantic: This can be:
`assert.equal(editor.range.head.section.text, 'Line1', 'can access section text');`

bantic: And the `{range}` second argument on the previous line can then go away.

eguitarz:
@bantic Thanks, now I see what you mean. Since there's no other changes, please close this PR. I agree with you that we don't  pass the editor in.

bantic:
@eguitarz Thanks, yes, I'll give this a close. And I would like to apologize for all of the back-and-forth here. Somehow I mixed this up with your initial PR to add the `willHandleNewline` hook  and failed to realize that this wasn't the same — for some reason I thought this was continued discussion about the synthetic event that we had discussed in that PR. I could have saved you a lot of time by recognizing that and then pointing out that the `editor` and `range` are already accessible without being passed...my apologies! 🙇 😿

Issue #498, opened 2016-09-16
Pasting into an empty Mobiledoc via the browser File menu does not work


YoranBrondsema:
Working on a fix, will be coming in a short moment.

bantic:
closed in #499

PR #499, opened 2016-09-16
Fix bug #498:  Pasting into an empty Mobiledoc via the browser File menu does not work

YoranBrondsema:


bantic:
thanks @YoranBrondsema !

Issue #502, opened 2016-10-05
More intelligent undo system


disordinary:
Currently every change in the post is placed on the undo stack, because of memory restrictions that stack is restricted to five.  This means that a user can only undo the last five characters typed.

Usually multiple events of the same type group together as one object on the undo stack.

Events which effect undo and redo can broadly be grouped into the following (and there are likely more):
1. Text entry events
2. Forward deletion events
3. Backward deletion events
4. Paste events
5. Card change events
6. Selection / cursor change events
7. Undo and Redo events

Transitioning from one event type to another creates a single element on the undo stack. 

So, my expectation as a user is to type a run of text that counts as one item on the undo stack. That is until I make a selection change, delete a character, or (potentially) a certain time limit passes, etc.

We can further group the above into different categories of undo and redo behavior:
- Events 1 -3 combine into a run to create one undo object. (a typed run, multiple backspaces to delete a word, multiple deletions to do the same).
- Events 4  and 5 always create a unique undo object and never go into a run but breaks the undo run of events 1 - 3.
- Event 6 never undos, but breaks the undo run of events 1 - 3.
- Event 7 obviously can undo and redo, and breaks the undo run of events 1 - 3.

---

I can think of three ways to accomplish this. 
1. Add every change onto the undo stack, but keep track of the first element in that run, if the run is broken then discard every item associated with that run except for the last one, so we only keep the final state and the state before the change begun.
2. Create a separate stack for run characters (similar to how you already store the `pendingSnapshot`) and push only the latest version onto the proper undo stack when the run is broken or an undo requested. 
3. Overwrite the last element of the undo stack with the `pendingSnapshot` if it's part of the same run.

Obviously the first two options would require a considerable increase in the number of states which are kept in memory. 

Is there a reason why the number of five was chosen as a default?

mixonic:
I *think* option three is reasonable @disordinary. But if you investigate and realize otherwise I would welcome some insight.

Five is fairly arbitrary, and it should also be configurable via `undoDepth` as a config option to the editor. We could come other with other metrics for the cost, such as a measurement of document complexity as a proxy for memory weight and make the cap work that way.  Presume we can only store N sections for undo, and if a new snapshot has Y sections that push the number of total in the stack over N then evict the oldest snapshot.

bantic:
Option 3 also seems reasonable to me. I don't think this would cause any issues.

[...]

PR #507, opened 2016-10-19
Group Undos

disordinary:
Refs #502

Basically there are three types of undo events, `insert`, `delete`, and everything else (which is undefined here).

`Insert` and `delete` events group together overwriting the last element in the undo queue when they are repeated unless another event occurs which breaks it or a timeout occurs.

So, if I write text then as long as I don't delete text or insert an atom or card, and as long as the timeout doesn't occur, those events are grouped by replacing the last element on the undo queue and are popped in one go.

A timeout is reset whenever the `run` method is called on the editor, so it doesn't happen on the first occurrence of an event in a run but rather the last.

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -43,6 +43,7 @@ const defaults = {
   spellcheck: true,
   autofocus: true,
   undoDepth: 5,
+  undoBlockTimeout: 5000, // ms for an undo event
   cards: [],
   atoms: [],
   cardOptions: {},
@@ -135,7 +136,7 @@ class Editor {
     this.post = this.loadPost();
     this._renderTree = new RenderTree(this.post);
 
-    this._editHistory = new EditHistory(this, this.undoDepth);
+    this._editHistory = new EditHistory(this, this.undoDepth, this.undoBlockTimeout);
     this._eventManager = new EventManager(this);
     this._mutationHandler = new MutationHandler(this);
     this._editState = new EditState(this);
@@ -687,7 +688,7 @@ class Editor {
     if (postEditor._shouldCancelSnapshot) {
       this._editHistory._pendingSnapshot = null;
     }
-    this._editHistory.storeSnapshot();
+    this._editHistory.storeSnapshot(postEditor.editActionTaken);
 
     return result;
   }
--- a/tests/acceptance/editor-undo-redo-test.js
+++ b/tests/acceptance/editor-undo-redo-test.js
@@ -64,12 +64,11 @@ test('undo/redo the insertion of a character', (assert) => {
 // when typing characters
 test('undo/redo the insertion of multiple characters', (assert) => {
   let done = assert.async();
-  let beforeUndo, afterUndo1, afterUndo2;
+  let beforeUndo, afterUndo;
   editor = Helpers.mobiledoc.renderIntoAndFocusTail(editorElement, ({post, markupSection, marker}) => {
     beforeUndo = post([markupSection('p', [marker('abcDE')])]);
-    afterUndo1 = post([markupSection('p', [marker('abcD')])]);
-    afterUndo2 = post([markupSection('p', [marker('abc')])]);
-    return afterUndo2;
+    afterUndo = post([markupSection('p', [marker('abc')])]);
+    return afterUndo;
   });
 
   let textNode = Helpers.dom.findTextNode(editorElement, 'abc');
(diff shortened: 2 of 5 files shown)


src/js/editor/editor.js
```diff
@@ -70,6 +70,16 @@ const CALLBACK_QUEUES = {
...
 
+// There are only two events that we're concerned about for Undo, that is inserting text and deleting content.
+// These are the only two states that go on a "run" and create a combined undo, everything else has it's own
+// deadicated undo.
+const EDITOR_RUN_LAST_ACTION_STATE = {
+  INSERT_TEXT: 1,
+  DELETE: 2
+};
+
+const UNDO_RUN_BREAK_TIMEOUT_LENGTH = 500; // how many ms after the last event before an UNDO run is broken
```
bantic: Mentioned in gitter: Increasing this seems worthwhile. 5000 ms is a good default. I'm wondering if it makes sense to allow it to be customizable, also...

src/js/editor/editor.js
```diff
@@ -293,7 +306,7 @@ class Editor {
     this.run(postEditor => {
       let nextPosition = postEditor.deleteAtPosition(position, direction, {unit});
       postEditor.setRange(nextPosition);
-    });
+    }, EDITOR_RUN_LAST_ACTION_STATE.DELETE);
```
bantic: I am reluctant to modify the method signature of the already-public `run` method in order to add a private-ish second parameter...

bantic: Also, there isn't necessarily anything about the `run` method that requires it to be only one of ["insert","delete"] — it's totally possible to do an `editor.run` that inserts at some places and deletes at others.
Not sure how important this distinction is. The most common use case is certainly for a `run` to focus on either one or the other.

src/js/editor/editor.js
```diff
@@ -673,10 +686,32 @@ class Editor {
...
+      clearTimeout(this._undoRunTimeout);
+      this._undoRunTimeout = null;
+    }
+
+    if(isInARun) {
+        this._undoRunTimeout = setTimeout(() => {
+          this._undoRunTimeout = null;
+          this._undoRunLastState = null;
+        },
+        UNDO_RUN_BREAK_TIMEOUT_LENGTH);
```
bantic: `this._undoRunTimeout` probably needs to be explicitly cleared in `editor#destroy` as well (if it is present)

tests/acceptance/editor-undo-redo-test.js
```diff
@@ -99,6 +92,46 @@ test('undo/redo the insertion of multiple characters', (assert) => {
   });
 });
 
+test('adding and deleting characters break the undo group/run', (assert) => {
```
bantic: Adding a test for the timeout functionality would also be good here (and is another reason that it could/should be configurable — at least for tests).

src/js/editor/editor.js
disordinary: I think if it's mixed then it's unlikely to be the result of a direct manipulation by the end user and therefore should break the undo grouping. The same as some events which are insertions like a Paste event.

The reason that this works is because only two states are explicitly defined (TEXT_INSERT and DELETE which results in a grouping) and everything else is undefined which keeps them as unique elements on the undo stack.

tests/acceptance/editor-undo-redo-test.js
disordinary: Yes,  or in the `undo/redo the insertion of multiple characters`  test above it.

Making the timeout configurable is a good option although I don't see people changing it too often if we come up with a timeout that works well.

src/js/editor/editor.js
disordinary: Yes, I agree with the private argument on run. Alternatives include:
- Having a more complicated return type that can include the undo grouping information.
- Duplicating Run which focuses on those two specific use cases.

bantic: @disordinary Sorry to let this linger — we were at EmberFest.
Yeah, hiding the second arg with a private `_run` totally makes sense to me. 👍

bantic:
@disordinary Thanks again for your work on this. If you want to add the private `_run` as described and a test that exercises the timeout functionality (which I imagine necessitates making the timeout configurable), I would be happy to merge!

disordinary:
@bantic, great I'll get a PR together by Monday.

disordinary:
I've updated the PR as follows:
- Introduced a new private `_run` command.
- Set the timeout for the undo blocks as an option `option.undoBlockTimeout` with a default of 5000 seconds (feel free to rename, you know that quote about the two hard things in computer science).
- Introduced new tests for the undo grouping timeout.

disordinary:
@bantic, I just noticed the failing tests. It seems to be a problem with the browsers themselves not being able to connect to the test environment and not anything to do with this PR, do you mind running again?

mixonic:
@disordinary I restarted them. Force pushing a no-op rebase would also kick them off. 👍

mixonic:
@disordinary I'll look closely in just a few hours here.

src/js/editor/editor.js
```diff
@@ -677,6 +695,45 @@ class Editor {
...
+    let isInARun = actionType && this._undoRunLastState === actionType;
+    this._undoRunLastState = actionType;
+
+    if(this._undoRunTimeout) {
+      clearTimeout(this._undoRunTimeout);
+      this._undoRunTimeout = null;
+    }
+
+    if(isInARun) {
+        this._undoRunTimeout = setTimeout(() => {
```
mixonic: This debouncing pattern isn't free- setting and clearing timeouts can be very slow JS code. Better might be to schedule a flush of this state in the future, and at the time of the flush check for if there have been any more recent actions that mean the flush should be cancelled. Then, if the flush is cancelled, schedule a new followup. This would be move the timeout setup/cancel work outside of the input and interactive parts of the flow.

src/js/editor/editor.js
```diff
@@ -677,6 +695,45 @@ class Editor {
...
+   *
+   * @param {Function} callback Called with an instance of
+   *        {@link PostEditor} as its argument.
+   * @param {Number|undefined} actionType provides the type of the current
+   *                                      action, but only if it's a "runnable"
+   *                                      type.
+   * @return {Mixed} The return value of `callback`.
+   * @private
+   */
+  _run(callback, actionType) {
```
mixonic: Does it feel wrong that the actionType stuff happens at this level? Wouldn't it be better to just have `deleteRange`, for example, set the state of lastState and schedule a some further work into the loop? That would seem to generically handle a lot of manual flagging of runloops, and to be more friendly to user-space code.

disordinary: The reason that the state is being passed as an argument into the `run` method and not being set on the object is that `run` somehow needs to know that the command that it's executing is the command that set the state, because any event other than `insertText`, `deleteAtPosition`, and `deleteRange` should default to breaking the undo grouping.  I can't think of an elegant solution off the top of my head unfortunately.

mixonic:
Excellent, after some pairing today this feels much improved and we're ready to land it pending a green build. Thanks @disordinary 👏 !

mixonic:
I'm merging this with a failing build. IE11 was failing previously due to the racy nature of the undo/redo tests (ripe for refactoring). By expanding the delay used I've gotten them passing locally. At this point the failing build is just Sauce not having the IE environment available to run against. Gonna move it into master and get it integrated with https://github.com/bustlelabs/mobiledoc-kit/pull/515 so we can look at a new point release.

PR #518, opened 2016-11-11
range#expandByMarker

mixonic:
This provides a high-level method to expand a range by checks on each marker. For example to expand the current cursor position to all markers with the same link:

```js
/* Collapsed means the cursor is not a selection */
if (editor.range.isCollapsed) {
  /* The marker (smallest unit markupable text) under the cursor */
  let marker = editor.range.head.marker;
  if (marker) {
    /* See if that marker has a link */
    let linkMarkup = marker.markups.find(m => m.tagName === 'a');
    if (linkMarkup) {
      /* If it has a link, expand the range to all markers with that link */
      let linkRange = editor.range.expandByMarker(marker => {
        return !!(marker.markups.find(m => m === linkMarkup);
      });
      /* Now you can use linkRange to do something to the whole link */
      editor.run(postEditor => {
        postEditor.removeMarkupFromRange(linkMarkup);
      });
    }
  }
}
```

TODO

* [x] ~~Should `expandRangeByMarker` be on the `section`? I think it probably should since the expansion is scoped to the section.~~ it will be on `range` instead. ~~`scanToMarker` will be on markers~~ derp this is just a detect with a step backwards. I've changed the expand logic to just use that.
* [x] Some further docs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 ---
 language: node_js
 node_js:
-  - "stable"
+  - "6"
 
-sudo: required
-dist: trusty
+sudo: false
 
 cache:
   directories:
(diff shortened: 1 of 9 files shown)


mixonic:
In 0.10.13

PR #524, opened 2016-11-16
Bump saucie and other deps

mixonic:


mixonic:
Saucie 2.0 seems to hang after sauce runs complete. Just moved back to 1.4.1.

PR #528, opened 2017-01-10
Keep empty sections in document

StephanHoyer:
Tests are failing atm. If this change is worth it, I might fix those tests.

### DIFF
--- a/src/js/parsers/dom.js
+++ b/src/js/parsers/dom.js
@@ -116,10 +116,6 @@ class DOMParser {
   }
 
   appendSection(post, section) {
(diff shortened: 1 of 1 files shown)


mixonic:
@StephanHoyer I'm not inclined to change the behavior without some reasoning- this is only during the parse step which is basically entirely a judgement call on how we expect default parsing to look. In theory anyone can write a custom parser to Mobiledoc that has different semantics.

Are there some reasons you want to share for advancing this?

StephanHoyer:
But changing the document that's gonna be edited? This it not what the user expects right?

PR #533, opened 2017-02-19
Fix the link to the github repo

erickmerchant:


bantic:
👍 thanks!

bantic:
@erickmerchant This actually needs to be fixed in master as well, because the gh-pages are built from master. I'll handle that.

PR #534, opened 2017-02-20
Allows for the unregistering of custom key commands.

disordinary:
- Allows users to add an optional `name` attribute to a keyCommand object as passed to the `registerKeyCommand` method.

- If the `name` attribute is added a new `unregisterKeyCommands` method will remove these key commands from the key command array.
Multiple key commands can have the same name, if so `unregisterKeyCommands` will remove all instances with that name.

Usage:

```javascript
  editor.registerKeyCommand({
    name: 'cut',
    str: 'ctrl+x',
    run(editor) { ... }
  });

  editor.unregisterKeyCommands('cut');
```

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -285,6 +285,20 @@ class Editor {
   }
 
   /**
+   * @param {String} name If the keyCommand event has a name attribute it can be removed.
+   * @public
+   */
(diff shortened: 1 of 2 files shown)


bantic:
+1 thanks @disordinary, looks good to me :)

PR #535, opened 2017-02-20
Replace deprecated mobiledoc-html-renderer with mobiledoc-dom-renderer

YoranBrondsema:


mixonic:
@YoranBrondsema this would not be safe for running in Node. I'm not sure we are 100% safe for that now but it would be nice to get closer and not further away.

The suggested way to render as HTML is to use the DOM renderer and SimpleDOM. [docs](https://github.com/bustlelabs/mobiledoc-dom-renderer#rendering-html).

Can we simply drop the `serializePost` options for `html` or `text`? Only permit serialization to Mobiledoc, and if you need to serialize into an alternative format just bring you own renderer. It would simplify our build and make mobiledoc-kit smaller to remove the dependency on mobiledoc-html-renderer and mobiledoc-text-renderer entirely.

@bantic?

bantic:
Being able to simplify by dropping mobiledoc-*-renderer as dependencies is appealing, for sure.

The reason we use [serializePost with 'html' and 'text' formats](https://github.com/bustlelabs/mobiledoc-kit/blob/c75711c151b1a31e80bc0f81a8f39f361756c569/src/js/editor/event-manager.js#L212-L214) is for copying out of an editor. We can't drop those formats without dropping support for external pasting of copied mobiledoc content.

YoranBrondsema:
@mixonic @bantic Does copy-pasting make sense in a Node.js environment? One application I can think of is in a Node.js-based desktop application such as the Atom editor. If so, I think we'd either have to have a user pass a DOM when initializing mobiledoc-kit or we add a dependency on SimpleDOM. If not, then we can assert beforehand that we must be in a browser-based environment to support copy-pasting.

bantic:
@YoranBrondsema I think you're right — copy/pasting doesn't make sense in a Node.js environment. Afaict event Atom or some other app running in electron will have access to a DOM. I am in favor of falling back to text serialization when there's not DOM available. I'll push up that change to this branch and get your thoughts.

bantic:
@YoranBrondsema What do you think of: https://github.com/YoranBrondsema/mobiledoc-kit/compare/yb-move-to-dom-renderer...bustlelabs:yb-move-to-dom-renderer ?

YoranBrondsema:
@bantic Looks great, no comments. Thanks a lot for making this get through!

bantic:
closed for #538

PR #538, opened 2017-03-09
Replace deprecated mobiledoc-html-renderer with mobiledoc-dom-renderer

bantic:
Fallback to text serialization when html not available

supersedes #535


bantic:
released in [v0.10.15](https://github.com/bustlelabs/mobiledoc-kit/releases/tag/v0.10.15)

PR #542, opened 2017-03-15
Constrain selection to editor element when probing for range

### LINKED ISSUE #486: Triple-clicking to select throws an error when editing is disabled

bantic:
This fixes an issue where `Position.fromNode` would be called with a node
that is outside the editor element, triggering a failed assertion.

Now, Cursor constrains the selection to only include the extent inside the
editor element before looking up Positions from the anchor and focus nodes.

Most of the time this situation is prevented by the browser (it refuses to
allow one to create a selection that crosses into or out of the
contentEditable div), but when `editor.disableEditing()` is called, a user
can triple-click the last part of the mobiledoc document which causes the
browser (for Chrome and Safari, but not Firefox) to extend the selection
*outside* the editor's element.

Fixes #486 as reported by @YoranBrondsema

### DIFF
--- a/assets/demo/debug.css
+++ b/assets/demo/debug.css
@@ -1,3 +1,22 @@
 #toolbar button.active {
   font-weight: bold;
 }
+
+#error {
+  color: red;
+}
+
+#error .name {
+  font-weight: bold;
+}
+
+#editor {
+  outline: 1px solid black;
+  width: 200px;
+  min-height: 300px;
+}
+
+#editor[contenteditable="false"] {
+  color: #9e9e9e;
+  background-color: #f1f1f1;
+}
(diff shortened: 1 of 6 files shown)


bantic:
released in [v0.10.16](https://github.com/bustle/mobiledoc-kit/blob/master/CHANGELOG.md#01016-2017-05-25)

PR #549, opened 2017-04-12
Fix minor typos in README

kevinansfield:
* fixes small things that jumped out when reading through


disordinary:
Thanks @kevinansfield, looks good!

Issue #551, opened 2017-04-14
Ctrl + U only works when text is selected


rawsh:
Underlining only works if text is selected, instead of working with the shortcut + typing, like bold and italic.

bantic:
fixed by #554

PR #554, opened 2017-04-27
Fix issue #551: "Ctrl + U only works when text is selected"

YoranBrondsema:


bantic:
released in [v0.10.16](https://github.com/bustle/mobiledoc-kit/blob/master/CHANGELOG.md#01016-2017-05-25)

PR #560, opened 2017-05-18
renaming bustlelabs -> bustle with the exception of gitter.im links

rhwlo:
# Summary

I changed the links pointing to `bustlelabs` to links pointing to `bustle`, with the exception of the `gitter.im` links.

## Motivation

I noticed that the name of your github org had changed and it had broken some links in various places. This seemed like a quick enough fix!

## Notes

I tried to be a little more intelligent than just a global find-and-replace; instead I matched specific URLs, and only one set of URLs (those from gitter.im) wouldn’t work with the name replaced. Hopefully this is helpful!

thanks for your consideration :smile:


bantic:
wow, thanks @rhwlo !

PR #573, opened 2017-08-17
Disable event manager when editing is disabled

### LINKED ISSUE #572: disableEditing allows for copy, paste, and cut
At Bustle we have writers accidentally pasting values into mobiledoc bodies that are created with `disableEditing` called.  The writers are then unable to delete the text; however, it can be removed with cut

Code:
We are using Ember Mobiledoc Editor `v0.5.12`, with settings `isEditingDisabled=true` in the template, which triggers [`disableEditing`](https://github.com/bustle/ember-mobiledoc-editor/blob/66d0e655ec95bc1a0f08ac2f1842802b2e913f63/addon/components/mobiledoc-editor/component.js#L258-L260)

Steps to reproduce:
- Create new mobiledoc editor with `disableEditing`
- Click on the body (no cursor should appear)
- Paste text
- Highlighting text and deleting does nothing
- Highlighting text and cut to remove text

bantic:
Previously, manual (as opposed to programmatic) editing was being
prevented for an editor after calling `#disableEditing` by setting
"contenteditable" back to false for the editor's element. This prevents the
browser from sending most of the edit-related events (keyup, keydown,
etc) that occur when editing manually. But some events (like paste) are
still emitted by the browser even though "contenteditable" is false.

This change calls `EventManager#stop` when editing is disabled, and
internally the event manager ignores all events when it is stopped.

Also some rearrangement

Other changes
  * Change editor's internal `isEditable` prop to default to true rather
    than null (incidentally allows a user to specify that an editor
    should start disabled by passing in that prop to the constructor)
  * Add test for triggering paste event in a disabled editor
  * add assertion `assert.isBlank`
  * better assertion actual/expected values for `hasElement` and `hasNoElement`
  * add test helper `getData` (for parity with the `setData` in
    element-utils and to avoid using jQuery's `.data()` method)
  * consolidate the editing-disabled tests into a new suite
  * make `editor#render` call `disableEditing` or `enableEditing` after
    it sets `hasRendered` to ensure that the event manager is stopped if
    the user calls `editor.disableEditing()` before calling
    `editor.render()`

Fixes #572

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -122,7 +122,7 @@ class Editor {
     assert('editor create accepts an options object. For legacy usage passing an element for the first argument, consider the `html` option for loading DOM or HTML posts. For other cases call `editor.render(domNode)` after editor creation',
           (options && !options.nodeType));
     this._views = [];
-    this.isEditable = null;
+    this.isEditable = true;
     this._parserPlugins = options.parserPlugins || [];
 
     // FIXME: This should merge onto this.options
@@ -239,10 +239,6 @@ class Editor {
 
     this.element = element;
 
-    if (this.isEditable === null) {
-      this.enableEditing();
-    }
-
     this._addTooltip();
 
     // A call to `run` will trigger the didUpdatePostCallbacks hooks with a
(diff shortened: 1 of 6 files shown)


bantic:
released in [v0.10.18](https://github.com/bustle/mobiledoc-kit/blob/master/CHANGELOG.md#01018-2017-08-17)

PR #579, opened 2017-08-31
upgrade deps

bantic:


YoranBrondsema:
@bantic Would it make sense to install something like [Greenkeeper](https://greenkeeper.io/) to avoid having to do this manually? I installed it for another project and it works really well.

bantic:
@YoranBrondsema Yeah, good idea. There's some issue with configuring greenkeeper for this repo (a permissions problem, maybe), but I emailed them to get it fixed.

YoranBrondsema:
@bantic Great, thanks!

Issue #581, opened 2017-09-10
Cursor is placed at the start of the line when typing special characters 


kiwicopple:
Note: using version 0.10.19

If the first letter in a new paragraph is a character like "#" or "-", after the user has typed the symbol the cursor is placed at the start.

### Steps to replicate 

1. In an editor, type the symbol "#"
2. Next, type a letter "a"

The expected output in the editor should be "#a", however it is actually "a#"

YoranBrondsema:
@kiwicopple Thanks for reporting. I can reproduce the bug and it seems to be caused by my own changes in #575. Looking into it now!

YoranBrondsema:
#582 reverts the changes and I added some more info there!

kiwicopple:
Thanks for the speedy fix @YoranBrondsema !

PR #575, opened 2017-08-18
Fix non-printable keys in firefox

YoranBrondsema:
Fixes #574

PR #582, opened 2017-09-10
Fix for issue #581

YoranBrondsema:
It turns out that `keyCode` is not the right way for uniquely identifying keys. For instance, `End` has a keycode of 35 but so does `Shift+3` (to make the character `#`).

This reverts changes by PR #575  but I suggest that we try to move to a more robust system for uniquely detecting keys. The changes by PR #575 only affected Firefox so I think we can safely use [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) at those locations. For the `End` key, `event.key === 'End'` and for `#`, it's `event.key === '#'`.

Maybe it's an idea to use `event.key` when available (I'm guessing it will be `undefined` for older browsers) and fall back on `event.keyCode`?

Curious to know what you think @bantic .

### DIFF
--- a/src/js/utils/key.js
+++ b/src/js/utils/key.js
@@ -79,8 +79,7 @@ const Key = class Key {
     if (this.isTab()) { return TAB; }
     return String.fromCharCode(this.charCode);
   }
-
-  isEscape() {
+isEscape() {
     return this.keyCode === Keycodes.ESC;
   }
 
@@ -125,7 +124,10 @@ const Key = class Key {
   }
 
   isEnd() {
-    return this.keyCode === Keycodes.END;
+    return (
+      this.keyCode === Keycodes.END &&
+      !this.modifierMask
+    );
   }
 
   isPageUp() {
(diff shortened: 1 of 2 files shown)


YoranBrondsema:
Also, looks like there's a problem with the SauceLabs integration :/

````
Error: Please, provide the username and the access key for SauceLabs. Use the option --help for more information.
````

bantic:
@YoranBrondsema Thanks for addressing this so quickly! Using `event.key` definitely looks like the better way, and MDN appears to agree:
image

I restarted the build. I thought the Travis-ci + Sauce integration was working as of #578 but maybe I need to revisit it. If the build fails again I'll take a look when I have some time.

YoranBrondsema:
@bantic Cool. Can we maybe merge this and make a new release? I think issue #581 is more serious than the bug that was fixed by #575 and that gives us some more time to work on a proper solution that uses `KeyboardEvent.key`.

YoranBrondsema:
Closed as it's superseded by #587

Issue #574, opened 2017-08-18
Pressing certain non-printable keys (Home, End,...) print a character on Firefox


YoranBrondsema:
The issue is similar to #565.

PR #587, opened 2017-10-04
Prefer KeyboardEvent.key over KeyboardEvent.keyCode

YoranBrondsema:
* Uses [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property if it exists (see https://caniuse.com/#feat=keyboardevent-key for browser usage) and uses [KeyboardEvent.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) only as a fallback as it's deprecated. This change addresses both #581 and #574.
* Small refactoring to not expose the `SPECIAL_KEYS` array but rather expose a method `specialCharacterToCode`. The idea is to keep the logic as much as possible inside the `Key` class in order to keep the key/keyCode stuff hidden.


YoranBrondsema:
@bantic I think so too, thanks for having a look at it!

YoranBrondsema:
@bantic Would you be able to create a new release to get this into it?

bantic:
@YoranBrondsema released in 0.10.20. Thank you!

PR #588, opened 2017-10-08
fix(docs): incorrect Marker definition signature

reconbot:
The text and example didn't seem to match.


YoranBrondsema:
Thank you!

PR #591, opened 2017-11-26
Split shortcuts by platform and name them

hpohlmeyer:
The main task of this pull request is to name the default shortcuts.

I think the default shortcuts are not bad, but if you want to use mobiledoc kit as a framework you need a way to reset the shortcuts and not just overwrite them. (For example, if you do not want to support underlined text)

You already have the `unregisterKeyCommands()` function to delete key commands. Unfortunately, the shortcuts need a name property to delete them. This is what I did for the default commands.

Some of them had different functions, depending on the platform. This is why I split the arrays into a mac and a windows array. This way I could give them distinctive names.

### DIFF
--- a/src/js/editor/key-commands.js
+++ b/src/js/editor/key-commands.js
@@ -38,101 +38,115 @@ function deleteToEndOfSection(editor) {
   });
 }
 
-export const DEFAULT_KEY_COMMANDS = [{
+const MAC_KEY_COMMANDS = [{
   str: 'META+B',
-  run(editor) {
-    editor.toggleMarkup('strong');
-  }
-}, {
-  str: 'CTRL+B',
+  name: 'default-bold',
   run(editor) {
     editor.toggleMarkup('strong');
   }
 }, {
   str: 'META+I',
-  run(editor) {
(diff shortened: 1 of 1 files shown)


hpohlmeyer:
Another approach to this would be to change the `unregisterKeyCommands()` function to take the key command instead of a name. This way you do not even remember the name of the shortcut, but can delete the unwanted shortcut.

hpohlmeyer:
Travis CI fails, because some of the key commands don’t work anymore. For example `CTRL+B` worked on macs, but did the same as `CMD+B`.  All windows commands, that use `CTRL` as a modifier are now associated with `CMD` on mac.

Before I start modifing the test, I would like to know if you welcome these changes. Thanks!

mixonic:
@hpohlmeyer thanks for diving in here, I've made a note to try and follow up on Friday. 🙇

lukemelia:
I ran into a similar issue as @hpohlmeyer and am supportive of his proposed change.

hpohlmeyer:
Closing this as @lukemelia has a modernized version of this PR in #789.

PR #593, opened 2017-11-28
Fix issue #592: Selection is off when triple-clicking on Firefox

### LINKED ISSUE #543: Double-clicking a link in Firefox causes hasActiveMarkup('a') to report incorrect results
Maybe I'm not coding something correctly on my end. So here is the gist of what I'm doing:

```javascript
editor.cursorDidChange(() => {
  editor.run(() => {
    let hasActiveLink = editor.hasActiveMarkup('a');
    let hasSelectedRange = !editor.range.isCollapsed;
    this.set('enableLinkButton', hasActiveLink || hasSelectedRange);
  });
});
```

All is fine across browsers until I double-click a linked word in Firefox. The UI correctly highlights the entire word, but `hasActiveLink` is `false` only in that scenario.

The answer to this Stack Overflow question is probably the [closest thing that I could find to a possible explanation](http://stackoverflow.com/questions/19234988/firefox-how-to-get-selected-text-when-using-double-click) of what's going on. When I poke around in the Mobiledoc source, it seems to match what's described in the answer: the parent node ends up being the containing `p` element and not the `a`, so I think that means it's including the `a` in the selection.

### LINKED ISSUE #592: Selection is off when triple-clicking on Firefox
I found this out when trying to cut triple-clicked selected text:

mobiledoc

The cause of this is explained here: https://stackoverflow.com/a/21234837/1269194.

YoranBrondsema:
Fixes #592 . Might also fix #543 (I'm not 100% sure, I haven't tried it).

### DIFF
--- a/src/js/utils/cursor/position.js
+++ b/src/js/utils/cursor/position.js
@@ -409,6 +409,12 @@ Position = class Position {
             sectionOffset += postNode.length;
           }
           position = new Position(section, sectionOffset);
+        } else if (offset >= elementNode.childNodes.length) {
+
+          // This is to deal with how Firefox handles triple-click selections.
+          // See https://stackoverflow.com/a/21234837/1269194 for an
+          // explanation.
+          position = section.tailPosition();
         } else {
           // The offset is 0 if the cursor is on a non-atom-wrapper element node
           // (e.g., a <br> tag in a blank markup section)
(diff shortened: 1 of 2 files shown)


YoranBrondsema:
I see the build didn't pass but that's because of the SauceLabs thing.

YoranBrondsema:
@bantic or @mixonic any updates?

kevinansfield:
I can confirm this fixes the issue. @mixonic it would be great to have a second pair of eyes with more context 🙂

kevinansfield:
@mixonic friendly ping to get an OK on this fix? We've been using it in production at Ghost for a few months now and haven't run into any problems arising from this change

PR #598, opened 2018-01-23
chore(dependencies): `yarn upgrade`

dtetto:
Had to `yarn upgrade` in order to be able to dev locally (otherwise, `broccoli-slow-trees` fails `yarn start`).

ESLint started to complain about the `complexity` rule for `selection-util`’s `findOffsetInNode`, so I disabled the rule for that declaration only (its complexity is 21, could probably be refactored).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   - "6"
 
-sudo: false
+sudo: required
 dist: trusty
 
 addons:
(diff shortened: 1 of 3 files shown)


mixonic:
I pushed a commit here that should make Travis happy. I think there is currently something funky with Chrome and `sudo: false`.

bantic:
@dtetto Thanks for this PR, and sorry that we didn't fully follow through. The test errors were unrelated to this change but after I fixed them (in #612)  there were some conflicts and rather than passing the buck back to you to rebase I did a clean yarn upgrade in #614 and merged it. I hope you don't mind. Thanks for the attention here (and the patience)!

PR #599, opened 2018-01-23
[WIP] Link editing

dtetto:
Currently, link editing is only possibly by selecting all of the linked text, clicking the link button to remove the markup, then clicking the link button to add a fresh link. It is easy to accidentally “split“ a link this way, and gets in the way of our common use case where we want the old link URL so, e.g. we can Google a new source for content that has 404’d or, more commonly, fix minor typos.

After exploring several UI and code paths, I think what is best is to have an “Edit” button in the link tooltip (that shows the URL on hover). This way, “Link” toolbar buttons implementing `toggleLink` would maintain their current behavior (toggle the link markup off), and the URL-editing button would explicitly appear alongside its URL.

**I’m currently stuck, though,** on getting the editor to recognize any programmatic selection—as far as I can tell, I am pushing the right data through the right internals—the DOM selection works,
and the `SelectionChangeObservers` fire, for instance—but haven’t been able to find the right combination of `setRange` & friends on `editor` or `postEditor` to get Mobiledoc to know the cursor is in the right spot…any ideas?

### DIFF
--- a/assets/demo/demo.js
+++ b/assets/demo/demo.js
@@ -37,8 +37,12 @@ let activateButtons = (parentSelector, editor) => {
   $(`${parentSelector} button`).click(function() {
     let button = $(this);
     let action = button.data('action');
-    let arg = button.data('arg');
+    if (action === 'toggleLink') {
+      Mobiledoc.UI.toggleLink(editor);
+      return;
+    }
 
+    let arg = button.data('arg');
     editor[action](arg);
   });
 };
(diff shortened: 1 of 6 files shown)


sdhull:
So fwiw, the link editing interface I built for [our site](https://www.boundlesslabs.com) basically does this:
```js
// remove a link that has the cursor, or was clicked
const activeLink = this.editor.activeMarkups.find(markup => markup.tagName === 'a');
let range = this.editor.range;
if(range.isCollapsed) {
  range = range.expandByMarker(marker => marker.markups.includes(activeLink));
}
this.editor.run(postEditor => {
  postEditor.removeMarkupFromRange(range, markup);
});

```

And possibly the sneakiest thing I figured out for editing the `href`:
```js
const activeLink = this.editor.activeMarkups.find(markup => markup.tagName === 'a');
activeLink.attributes.href = newHref;
this.editor.run(postEditor => {
  postEditor._markDirty(this.editor.range.headMarker);
});
```
🙈 🙉 🙊

Issue #600, opened 2018-02-11
Copy-pasting from Word or LibreOffice Writer removes spaces between words


YoranBrondsema:
See [this comment](https://github.com/textAngular/textAngular/issues/598#issuecomment-92419636) for an explanation.

I think this issue comes from a single-character typo though :-). The solution in https://github.com/bustle/mobiledoc-kit/issues/333 mentions that newlines should be replaced with space characters. However, the actual implementation for the fix in https://github.com/bustle/mobiledoc-kit/commit/28ec7817a82720fc47d2a3aa704a0a143970955b replaces them with an empty string instead.

bantic:
fixed by #601

bantic:
released in v0.10.21 https://github.com/bustle/mobiledoc-kit/releases/tag/v0.10.21 thank you!

PR #601, opened 2018-02-11
Fix issue #600

YoranBrondsema:


bantic:
The test failure problems should be fixed on master; merging this in now and will deal with any test failures if they still occur. Thanks for this pr @YoranBrondsema !

PR #620, opened 2018-05-21
Fix error when pasting HTML that parses to a blank doc

### LINKED ISSUE #619: Paste with certain HTML throws `Cannot read property 'tailPosition' of undefined`
mobiledoc-kit throws an error when pasting the following HTML (tested in our app and the mobiledoc-kit demo):

```html
<meta charset='utf-8'><img src="http://ghost.blog/content/images/2018/05/Cosmos05.jpg" class="kg-image" style="box-sizing: inherit; margin: 1.5em auto; padding: 0px; border: 0px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-numeric: inherit; font-variant-east-asian: inherit; font-weight: 400; font-stretch: inherit; line-height: inherit; font-family: Georgia, serif; font-size: 22px; vertical-align: middle; max-width: 840px; display: block; color: rgb(60, 72, 78); letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-style: initial; text-decoration-color: initial;"><br class="Apple-interchange-newline">
```

The above HTML is what Chrome put into the clipboard when selecting and copying an image.

error:
```
post-inserter.js:262 Uncaught TypeError: Cannot read property 'tailPosition' of undefined
    at Visitor._replaceSection (post-inserter.js:262)
    at Visitor.value (post-inserter.js:55)
    at Visitor.visit (post-inserter.js:34)
    at Inserter.insert (post-inserter.js:348)
    at PostEditor.insertPost (post.js:1225)
    at event-manager.js:297
    at Editor.run (editor.js:750)
    at EventManager.paste (event-manager.js:296)
    at EventManager._handleEvent (event-manager.js:127)
    at HTMLDivElement.listener (event-manager.js:68)
_replaceSection @ post-inserter.js:262
value @ post-inserter.js:55
visit @ post-inserter.js:34
insert @ post-inserter.js:348
insertPost @ post.js:1225
(anonymous) @ event-manager.js:297
run @ editor.js:750
paste @ event-manager.js:296
_handleEvent @ event-manager.js:127
listener @ event-manager.js:68
```

kevinansfield:
closes #619
- it's possible to paste HTML that parses into a blank doc, if the paste results in a section being replaced then the `post-inserter` would throw an error because it expects the new post to have sections
- adds a guard in the paste event handler to avoid inserting a blank post

QUESTIONS:
- [ ] do we want to keep the guard in the paste event or should the fix actually be in `post-inserter.js`?
- [ ] are we happy that a paste which results in a blank doc still performs the deletion and empty markup section creation? (I'm assuming a paste that parses to a blank doc was not intentional so may be confusing if the user only sees content being deleted)

### DIFF
--- a/src/js/editor/post/post-inserter.js
+++ b/src/js/editor/post/post-inserter.js
@@ -278,7 +278,9 @@ export default class Inserter {
 
   insert(cursorPosition, newPost) {
     let visitor = new Visitor(this, cursorPosition);
-    visitor.visit(newPost);
+    if (!newPost.isBlank) {
+      visitor.visit(newPost);
+    }
     return visitor.cursorPosition;
   }
 }
--- a/tests/acceptance/editor-copy-paste-test.js
+++ b/tests/acceptance/editor-copy-paste-test.js
@@ -412,3 +412,25 @@ test('paste with shift key pastes plain text', (assert) => {
 
   assert.postIsSimilar(editor.post, expected);
 });
+
+test('paste with html that parses to blank doc doesn\'t error', (assert) => {
+  let expected;
+  let mobiledoc = Helpers.mobiledoc.build(({post, markupSection, marker}) => {
+    expected = post([
+      markupSection('p', [])
+    ]);
+
+    return post([
+      markupSection('p', [marker('abcd')])
+    ]);
+  });
+
+  editor = new Editor({mobiledoc, cards});
+  editor.render(editorElement);
+
+  Helpers.dom.setCopyData('text/html', `<div></div>`);
+  editor.selectRange(editor.post.toRange());
+  Helpers.dom.triggerPasteEvent(editor);
+
+  assert.postIsSimilar(editor.post, expected);
+});


jasonbekolay:
We have had this issue pop up a couple times in our app at Yapp and I believe this PR would improve the current experience. The second question does concern me a bit, but I'm not sure what would be better behavior. Having the editor do nothing in that case would not be a better experience in my opinion.

src/js/editor/event-manager.js
```diff
@@ -251,10 +251,12 @@ export default class EventManager {
     let targetFormat = this.modifierKeys.shift ? 'text' : 'html';
     let pastedPost = parsePostFromPaste(event, editor, {targetFormat});
 
-    editor.run(postEditor => {
-      let nextPosition = postEditor.insertPost(position, pastedPost);
-      postEditor.setRange(nextPosition);
-    });
+    if (!pastedPost.isBlank) {
+      editor.run(postEditor => {
+        let nextPosition = postEditor.insertPost(position, pastedPost);
```
mixonic: @kevinansfield I think we should have `insertPost` be able to insert a blank post. That seems valid.

kevinansfield: 👍

mixonic:
Awesome, thanks @kevinansfield!

PR #624, opened 2018-06-13
Add react renderer to readme

dfairaizl:
Hey guys - we've had this for a long time but never got around to adding it to the readme so here it is!


mixonic:
Awesome, thanks Dan!

Issue #625, opened 2018-06-15
Modifier keys get stuck when editor loses focus


kevinansfield:
We've had a number of users complain about <kbd>Backspace</kbd> deleting words rather than characters.

So far the only reliable reproduction I've found is when the <kbd>Alt</kbd> key is held down and the editor element loses focus, either by clicking outside of the editor or through a keyboard mapping that opens a window (eg, when Dash is mapped to <kbd>Alt + Space</kbd>). It would appear that the modifier keys in the `event-manager` are not reset when focus is lost.

This screen cap demonstrates the issue. Unfortunately it didn't capture the <kbd>Alt</kbd> key presses but it shows a single backspace working fine, pressing and holding the alt key whilst clicking off the editor before releasing the alt key, then afterwards a single backspace deleting whole words:

![sticky-alt-key](https://user-images.githubusercontent.com/415/41479226-65e3c344-70c2-11e8-84b6-035cf92eb5f4.gif)

PR #626, opened 2018-06-18
Fix sticky modifier keys for word deletion

kevinansfield:
closes https://github.com/bustle/mobiledoc-kit/issues/625
- use `key.altKey/shiftKey` getters when instigating word deletion in `event-manager.js`
    - word deletion is only triggered on <kbd>Backspace</kbd> and <kbd>Delete</kbd> so we can use the the browser-provided modifiers on the KeyboardEvent instead the internal modifier key tracking which is not reliable when the editor loses focus
- update `triggerDelete` and `triggerForwardDelete` test helpers to accept an `options` argument so that modifier keys can be passed in as event arguments

### DIFF
--- a/tests/acceptance/editor-sections-test.js
+++ b/tests/acceptance/editor-sections-test.js
@@ -5,6 +5,7 @@ import { NO_BREAK_SPACE } from 'mobiledoc-kit/renderers/editor-dom';
 import Range from 'mobiledoc-kit/utils/cursor/range';
 import Keycodes from 'mobiledoc-kit/utils/keycodes';
 import Browser from 'mobiledoc-kit/utils/browser';
+import { DIRECTION } from 'mobiledoc-kit/utils/key';
 
 const { test, module } = Helpers;
 
@@ -664,19 +665,23 @@ test('delete with option (Mac) or control (Win)  key deletes full word', (assert
 
   editor.selectRange(new Range(editor.post.tailPosition()));
 
-  let keyCode = Browser.isMac() ? Keycodes.ALT : Keycodes.CTRL;
+  let altKey, ctrlKey;
+  if (Browser.isMac()) {
+    /* Mac key codes for navigation by word */
+    altKey = true;
+    ctrlKey = false;
+  } else {
+    /* PC key codes for navigation by word */
+    altKey = false;
+    ctrlKey = true;
+  }
 
-  Helpers.dom.triggerKeyEvent(editor, 'keydown', {keyCode});
   Helpers.wait(() => {
-    Helpers.dom.triggerDelete(editor);
+    Helpers.dom.triggerDelete(editor, DIRECTION.BACKWARD, {altKey, ctrlKey});
 
     Helpers.wait(() => {
-      Helpers.dom.triggerKeyEvent(editor, 'keyup', {keyCode});
-
-      Helpers.wait(() => {
-        assert.postIsSimilar(editor.post, expected);
-        done();
-      });
+      assert.postIsSimilar(editor.post, expected);
+      done();
     });
   });
 });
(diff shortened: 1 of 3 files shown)


tests/acceptance/editor-sections-test.js
```diff
@@ -665,18 +666,15 @@ test('delete with option (Mac) or control (Win)  key deletes full word', (assert
   editor.selectRange(new Range(editor.post.tailPosition()));
 
   let keyCode = Browser.isMac() ? Keycodes.ALT : Keycodes.CTRL;
+  let altKey = Browser.isMac();
+  let ctrlKey = !Browser.isMac();
```
mixonic: Can you refactor this for clarity?

```js
let keyCode, altKey, ctrlKey;
if (Browser.isMac()) {
  /* Mac key codes for navigation by word */
  keyCode = Keycodes.ALT;
  altKey = true;
  ctrlKey = false;
} else {
  /* PC key codes for navigation by word */
  keyCode = Keycodes.CTRL;
  altKey = false;
  ctrlKey = true;
}
```

tests/helpers/dom.js
```diff
@@ -154,18 +154,17 @@ function createMockEvent(eventName, element, options={}) {
   return event;
 }
 
-function triggerDelete(editor, direction=DIRECTION.BACKWARD) {
+function triggerDelete(editor, direction=DIRECTION.BACKWARD, options={}) {
```
mixonic: Can we document what `options` is expected to be here? I presume `{ctrlKey: true}` and such.

tests/acceptance/editor-copy-paste-test.js
```diff
@@ -407,7 +407,7 @@ test('paste with shift key pastes plain text', (assert) => {
   Helpers.dom.triggerCopyEvent(editor);
   editor.selectRange(editor.post.tailPosition());
 
-  Helpers.dom.triggerKeyEvent(editor, 'keydown', { keyCode: Keycodes.SHIFT });
+  Helpers.dom.triggerKeyEvent(editor, 'keydown', { keyCode: Keycodes.SHIFT, shiftKey: true });
```
mixonic: This will update the modifier map on the event manager:

```js
Helpers.dom.triggerKeyEvent(editor, 'keydown', { keyCode: Keycodes.SHIFT });
```

But I don't believe this will:

```js
Helpers.dom.triggerKeyEvent(editor, 'keydown', { shiftKey: true });
```

Why add the `shiftKey` value?

kevinansfield:
> If we're trusting the browser to provide modifiers on the event for alt and control, why not also for shift? Is the event manager modifier map not needed at all?

The modifier map is still needed for shift because we look for the shift key being pressed inside the `paste` event and unfortunately the browser doesn't expose any keyboard information via the `PasteEvent` object.

tests/acceptance/editor-copy-paste-test.js
kevinansfield: Oops, that is a hangover from an earlier iteration I had where the modifier map was updated by looking at the `key.shiftKey` value on every keyup/keydown event.

tests/acceptance/editor-sections-test.js
kevinansfield: 👍

tests/helpers/dom.js
kevinansfield: 👍

mixonic:
@kevinansfield this is very clean after your updates, thanks. I reviewed the spec around the clipboard event and yeah, it doesn't seem to carry any data about modifiers. `ClipboardEvent` inherits from `Event`.

`altKey` for example is defined on `MouseEvent`: https://www.w3.org/TR/uievents/#dom-mouseevent-altkey and on `KeyboardEvent`: https://www.w3.org/TR/uievents/#dom-keyboardevent-altkey

I think it is an oversight of the spec that we don't have the same API for `ClipboardEvent` and need to synthesize it. I think with that in mind we could consider making the whole event manager a "shift modifier polyfill for paste events".

PR #627, opened 2018-06-18
Run parser plugins for top-level unknown elements

### LINKED ISSUE #494: parser plugins should parse top-level unknown and void elements
The Section parser first infers a section type for a given element before attempting to parse its child nodes, and it only runs its parser plugins against the child nodes. This means that HTML with root-level elements do not get run through the plugins. This happens (in slightly different ways) with both void elements and elements with childNodes:
- `parse('<video>...</video>')` — plugins skip `<video>`
- `parse('<video>')` — plugins skip  `video`

kevinansfield:
closes #494
- run parser plugins for the top-level section elements
    - do not continue parsing children elements if a parser marks the top-level element as finished
- update tests to double-check the output for top-level elements is unchanged

@bantic I'd be interested in a quick review and any pointers for any edge cases/tests you think may need adding 😄

### DIFF
--- a/src/js/parsers/section.js
+++ b/src/js/parsers/section.js
@@ -69,7 +69,13 @@ class SectionParser {
 
     this._updateStateFromElement(element);
 
-    let finished = this.runPlugins(element);
+    let finished = false;
+
+    // top-level text nodes will be run through parseNode later so avoid running
+    // the node through parserPlugins twice
+    if (!isTextNode(element)) {
+      finished = this.runPlugins(element);
+    }
 
     if (!finished) {
       let childNodes = isTextNode(element) ? [element] : element.childNodes;
--- a/src/js/parsers/section.js
+++ b/src/js/parsers/section.js
@@ -69,14 +69,18 @@ class SectionParser {
 
     this._updateStateFromElement(element);
 
-    let childNodes = isTextNode(element) ? [element] : element.childNodes;
+    let finished = this.runPlugins(element);
 
-    if (this.state.section.isListSection) {
-      this.parseListItems(childNodes);
-    } else {
-      forEach(childNodes, el => {
-        this.parseNode(el);
-      });
+    if (!finished) {
+      let childNodes = isTextNode(element) ? [element] : element.childNodes;
+
+      if (this.state.section.isListSection) {
+        this.parseListItems(childNodes);
+      } else {
+        forEach(childNodes, el => {
+          this.parseNode(el);
+        });
+      }
     }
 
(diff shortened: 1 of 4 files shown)


src/js/parsers/section.js
```diff
@@ -69,14 +69,18 @@ class SectionParser {
 
     this._updateStateFromElement(element);
 
-    let childNodes = isTextNode(element) ? [element] : element.childNodes;
+    let finished = this.runPlugins(element);
```
mixonic: Given an `element` here that is a text node, wouldn't the add call here mean that text node will be passed to `this.runPlugins` twice? Once here and once in `pardeNode`.

tests/unit/editor/editor-test.js
```diff
@@ -354,7 +354,7 @@ test('editor parses HTML post using parser plugins', (assert) => {
   editor = new Editor({html, parserPlugins: [parserPlugin]});
   assert.ok(!!editor.post, 'editor loads post');
 
-  assert.deepEqual(seenTagNames, ['TEXTAREA', 'IMG']);
+  assert.deepEqual(seenTagNames, ['P', 'TEXTAREA', 'IMG']);
```
mixonic: Is this characterized as a bugfix or as a breaking change?

src/js/parsers/section.js
kevinansfield: Good catch 😄 Pushed a fix

tests/unit/editor/editor-test.js
kevinansfield: I would characterise it as a bugfix because the old behaviour meant elements that were expected to go through parser plugins were never seen. I suppose it may be a breaking change if a consumer has built parser plugins to work around the buggy behaviour, I haven't been able to think up an example of that yet 🤔

mixonic: We talked about this and considered some examples. It seems like it wouldn't break any code we looked at in our quick review, so I'm going to call this a bugfix 👍

sdhull:
Looking forward to trying `parserPlugins` with this update, thanks @kevinansfield!! 

We found that when our users pasted in `<b>` or `<i>` tags, the could not un-bold or un-italicize that text because our toolbar toggles `strong` and `em` tags. I tried to add parser plugins that would convert `b` and `i` tags to `strong` and `em` respectively however I found that my plugins were never called for these tags.

I've worked around this by using computed properties (eg toggle `boldTag` instead of `"strong"`) in Ember but it's not an ideal solution imo.

PR #635, opened 2018-07-30
Fixed Ctrl+Backspace word delete on Linux

### LINKED ISSUE #634: Ctrl+Backspace does not delete word on Linux
<kbd>Ctrl+Backspace</kbd> on Linux has the same default OS behaviour of deleting a word as on Windows but mobiledoc-kit only deletes the last character.

kevinansfield:
closes #634
- toggle <kbd>Ctrl+Backspace</kbd> behaviour based on not being on Mac rather than checking for Windows explicitly


mixonic:
Thanks @kevinansfield!

Issue #445, opened 2016-08-03
add delete-to-start/end-of-line


bantic:
On Mac, pressing <kbd>meta</kbd>+<kbd>delete</kbd> deletes from the cursor position to the start of the line. We would like to have this work properly in mobiledoc-kit as well.
Need to:
  * [ ] add positioning heuristics to detect the position in a section that matches the start or end of its visual line (see [Position#atPoint](https://github.com/bustlelabs/mobiledoc-kit/blob/f91949069d4ead548a420f9810dc69cbce58a5eb/src/js/utils/cursor/position.js#L83-L94) for an example of where we currently do something similar)
  * [ ] listen for delete key with <kbd>meta</kbd> also pressed in [EventManager](https://github.com/bustlelabs/mobiledoc-kit/blob/0cc11bedc25f2746d702a402909553490b9b460f/src/js/editor/event-manager.js#L162-L172)
  * [ ] update `Position#move` to handle a `"line"` unit

see #403

kevinansfield:
Similar to [my comment here](https://github.com/bustle/mobiledoc-kit/issues/628#issuecomment-403857270) regarding the <kbd>alt</kbd>+<kbd>backspace</kbd> behaviour, removing the [custom delete key handler](https://github.com/bustle/mobiledoc-kit/blob/master/src/js/editor/event-manager.js#L182-L193) allows <kbd>meta</kbd>+<kbd>delete</kbd> to function.

Perhaps we can find a solution that leverages the default browser behaviour and mobiledoc-kit's mutation observer rather than trying to re-implement OS behaviour?

mixonic:
@kevinansfield I'm actually looking at being able to operate without the keypress events as part of a solution for https://github.com/bustle/mobiledoc-kit/issues/589.

We use key events for a few reasons:

* Our renderer has not really been a full diffing engine. When any marker in a section has changed, we simply re-render all the contents of the section. My work right now is to make that part of the render more detailed in its tracking. This should allow us to re-parse and then re-use the DOM a bit more often.
* Some browsers don't inject characters the same way. I don't have an example right now, but an ok thought-experiment would be to consider what happens when a user hits `enter`. Do we get a `<br>` in the `<p>`? Or a new `<p>`? After we re-parse we may need to correct the DOM, which is possibly just tricky.
[...]

PR #636, opened 2018-07-31
[SPIKE] Add delete to start of line for MacOS

kevinansfield:
refs https://github.com/bustle/mobiledoc-kit/issues/445
- add `Position.atStartOfLine` method that takes an existing position from which to find the start of it's line
- add `line` unit for use in `editor.performDelete`
- update delete key event handler to pass `line` unit when <kbd>Meta+Backspace</kbd> is pressed

I had a quick stab at re-implementing the default MacOS <kbd>Cmd+Backspace</kbd> behaviour but it's highlighted a problem. macOS allows you to place the caret visually at the end of the line when it's actual position is at the start of the next line (either through clicking at the end of the line or using <kbd>Cmd+Right</kbd>). You can see a demonstration here:

delete-line-buggy

When using both mobiledoc-kit's internal `position` and the browser-provided `document.getSelection().getRangeAt(0).getBoundingClientRect()` the only position exposed to JS is the beginning of the next line. The reason this is a problem is because to faithfully replicate macOS' default behaviour the delete needs to happen from the visual caret position rather than the logical caret position.

I disabled the delete key event handling in `event-manager.js` to demonstrate the native macOS behaviour where you can see the delete happening from the visual rather than logical position:

delete-line-native

### DIFF
--- a/src/js/editor/event-manager.js
+++ b/src/js/editor/event-manager.js
@@ -184,6 +184,8 @@ export default class EventManager {
         let unit = 'char';
         if (key.altKey && Browser.isMac()) {
           unit = 'word';
+        } else if (key.metaKey && Browser.isMac()) {
+          unit = 'line';
         } else if (key.ctrlKey && Browser.isWin()) {
           unit = 'word';
         }
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -401,7 +401,16 @@ class PostEditor {
       this.toggleSection('p', position);
       return this._range.head;
     } else {
-      let prevPosition = unit === 'word' ? position.moveWord(BACKWARD) : position.move(BACKWARD);
+      let prevPosition;
+
+      if (unit === 'word') {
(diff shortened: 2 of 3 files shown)


kevinansfield:
Unfortunately removing the delete key handler causes problems as soon as cards are involved - Cmd+Backspace can remove part of the card's DOM leaving the editor in a broken state - so it looks like a solution similar to this PR is needed at least in the short term.

Looking at other editor implementations I saw the following behaviour:

- Dropbox Paper: as per this PR
- Google Docs: works like native (they have their own canvas which re-implements everything related to cursor positioning so that’s expected)
- Medium: deletes to the beginning of the paragraph :face_with_raised_eyebrow:
- Quill: works like native (built around mutation observers and no complex cards to deal with so that’s expected)
- TinyMCE: works like native but undo breaks

mixonic (changes requested):
I think this is an excellent direction to take this @kevinansfield!

Issue #637, opened 2018-08-01
Cutting doesn't capture card section immediately following a list section


kevinansfield:
When cutting content that includes a card section immediately following a list section, the card is not captured as part of the serialised mobiledoc in the clipboard's `html` attribute.

![mobiledoc-cut-list-card](https://user-images.githubusercontent.com/415/43525950-f782da72-959a-11e8-91c3-e18e58a59cd1.gif)

Example `html` when a non-card section separates the list and card sections:

```
<meta charset='utf-8'><div data-mobiledoc='{"version":"0.3.1","atoms":[],"cards":[["markdown",{"markdown":"markdown card"}]],"markups":[],"sections":[[3,"ul",[[[0,[],0,"list"]]]],[1,"p",[[0,[],0,"test"]]],[10,0],[1,"p",[[0,[],0,"test"]]]]}'><div><ul><li>list</li></ul><p>test</p><p>test</p></div></div>
```

Example `html` when a card section immediately follows a list section:

```
<meta charset='utf-8'><div data-mobiledoc='{"version":"0.3.1","atoms":[],"cards":[],"markups":[],"sections":[[3,"ul",[[[0,[],0,"list"]]]],[1,"p",[[0,[],0,"test"]]]]}'><div><ul><li>list</li></ul><p>test</p></div></div>
```

kevinansfield:
Initial testing appears to suggest the problem is in `post.trimTo(range)`. I'm looking at creating a failing test case and fix.

PR #638, opened 2018-08-01
Fixed copy/paste of card section following a list section

kevinansfield:
closes https://github.com/bustle/mobiledoc-kit/issues/637
- if a non-markerable section immediately followed a list section the `post.trimTo()` was not appending it to the post
- resetting the `sectionParent` to the `post` object when hitting a non-markerable section allows all sections to be appended to the constrained post


mixonic:
This looks correct, good hunting @kevinansfield!

Issue #621, opened 2018-05-31
Spellcheck underlines flicker when typing


kevinansfield:
When typing in a markup section any spellcheck underlines will flicker on every character change which can be rather distracting. This is possibly related to https://github.com/bustle/mobiledoc-kit/issues/522 but I haven't noticed slowdown, just the flickering.

Examining other editors such as Medium, Dropbox Paper, and Quill, it would seem they do not suffer the same problem because they are not re-rendering when typing. Quill is the most well behaved out of the ones I looked at, both Medium and Paper flicker on <kbd>Backspace</kbd>, Medium also flickers on <kbd>Space</kbd>.

I had a look into how to re-use elements in [editor-dom](https://github.com/bustle/mobiledoc-kit/blob/master/src/js/renderers/editor-dom.js) rather than re-rendering with new elements. I think there are a couple of issues that need to be resolved to get rid of the flickering:

- [markup sections are always re-rendered](https://github.com/bustle/mobiledoc-kit/blob/master/src/js/renderers/editor-dom.js#L335-L337)
- when typing, `_joinSimilarMarkers` marks the existing Marker (and therefore TextNode) for deletion. Removal happens before the render which means it's not possible update the TextNode's `nodeValue` in the [markup renderer](https://github.com/bustle/mobiledoc-kit/blob/master/src/js/renderers/editor-dom.js#L377-L391) but a completely new TextNode has to be generated and added to the parent Marker or MarkupSection

I'm happy to dig into this but I'm interested to hear if there are any recommended approaches.

QUESTIONS:
- [ ] [this comment](https://github.com/bustle/mobiledoc-kit/blob/master/src/js/renderers/editor-dom.js#L335-L336) suggests that attributes on a markup section element may change, I could only see a way for the tag name or class name to be changed. Have I missed something?

kevinansfield:
On further testing it would appears the flickering is actually caused by changing the window selection.

Each selection change triggers Chrome's spellcheck to run resulting in the flicker as can be seen in this code pen https://codepen.io/lookingsideways/pen/gKpWPO?editors=1010 (As an aside, this pen also highlights how the selection change breaks spellcheck behaviour in Safari).

[...]

Issue #589, opened 2017-10-23
Input on android has a bug


part145pro:
Android:
When inserting text in the editor, after the first char the cursor moves back to start of editor area (instead than remaining after the inserted char), and after inserting another char the first char is duplicated.
------------------
the cursor is shown like '|'
the area delimiter is '<<' and '>>'
-----
bug :
-----
1) text area: <<>>
2) action: <press char 'a'>
3) text area: <<|A>>
4) action: <press char 'b'>
5) text area: <<Ab|A>>
.. and so on

raycohen:
I've observed the same bug pointing my android browser to http://bustle.github.io/mobiledoc-kit/demo/ as well as in our app. Its severe enough to make it borderline unusable. Any ideas on how to fix?

mixonic:
@raycohen Thanks for the reproduction.

bantic:
@raycohen and @part145pro Thanks for raising this issue. Does this screengrab from an emulator accurately show what you are seeing? I notice a few issues related to selection, and it seems that after the first character that I type, the remaining characters seems to function ok. I agree that this should be fixed, though. (The text I typed is "a" then "b" then "c").

![kapture 2018-06-01 at 16 29 15](https://user-images.githubusercontent.com/2023/40862110-691bfedc-65b9-11e8-850b-c8019db8080e.gif)

raycohen:
@bantic Yes, that screengrab matches the behavior I get in Chrome on a Pixel 2.

Firefox on android is definitely less important to us, but it has an additional quirk - the first keypress 'a' - gets swallowed and nothing appears. So if you type `a b c d` you'd end up seeing `cd[cursor]b`, and the cursor would only initially appear _after_ pressing `a`.

mixonic:
I've been looking at this a bit. What I've dug up so far:

* First off Chrome for Android does not trigger `keypress`: https://bugs.chromium.org/p/chromium/issues/detail?id=118639 This is intentional. There is no key being pressed, I believe is the logic. `keyCode` values are also not present on `keyup` and `keydown`. Key codes pertain to keys pressed on a physical keyboard, and again there is no physical keyboard.
[...]

PR #641, opened 2018-08-15
[SPIKE] Fix typing on Android

whatthewhat:
This is not a proper fix for https://github.com/bustle/mobiledoc-kit/issues/445, but rather a hacky workaround for anyone that really needs some kind of Android support. It does however work on my Pixel and in the emulator without breaking existing use cases. I'll keep this branch alive until a better fix is implemented.

[Screen recording of this branch on an actual device](https://www.dropbox.com/s/xtugf75xbq01oj7/mobiledoc.mp4?dl=0)

After spending a bit of time reading code I'm still not familiar enough with Mobiledoc internals to attempt a proper fix.. @mixonic I wonder if your attempts to operate without keypress ([1](https://github.com/bustle/mobiledoc-kit/issues/445#issuecomment-403903643), [2](https://github.com/bustle/mobiledoc-kit/issues/589#issuecomment-399323579)) were sucessfull? That approach could probably fix https://github.com/bustle/mobiledoc-kit/issues/621 as well by eliminating renderers?

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -462,7 +462,7 @@ class Editor {
     });
     this.rerender();
     if (currentRange) {
-      this.selectRange(currentRange);
+      this.selectRange(currentRange.move(DIRECTION.FORWARD));
     }
 
     this.runCallbacks(CALLBACK_QUEUES.DID_REPARSE);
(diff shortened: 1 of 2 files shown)


whatthewhat:
Updated branch to the latest mobiledoc-kit

Issue #644, opened 2018-08-22
Cards payload quotations and the standard


randohinn:
Hello. I'm writing a MobileDoc renderer in PHP. However, php really does not like to decode from json, when keys do not have quotation marks around them... Eg, this fails if I don't surround `src` in quotes.

```json
  "cards": [
    ["image", {
        "src": "http://google.com/logo.png"
    }]
],
```

I was wondering, is the quotation marks something one can configure in the kit output, or how much "in-spec" they are with the MobileDoc standard? If these are not configurable, I suggest making it so they are 😄.

randohinn:
Found workaround on PHP side

mixonic:
@randohinn All Mobiledocs should be valid JSON. If it ain't JSON, it ain't Mobiledoc.

If you're seeing something to the contrary, please let me know!

randohinn:
@mixonic Yup, the cards examples in mobiledoc.md, which describes the format, and iirc, also the atoms section there has unquoted examples.

randohinn:
Reopening as this JSON fault still exists in the docs. https://github.com/bustle/mobiledoc-kit/blob/master/MOBILEDOC.md under Card definition signature.

mixonic:
Closed by https://github.com/bustle/mobiledoc-kit/pull/645, feel free to re-open if I am missing something.

randohinn:
This has not been automatically closed, apparently. There are some more JSON errors in the examples, but I'll have to pass on fixing them, on account of not having the time right now. However, I'd be really happy, if someone could point out all the unsupported features in [my renderer](https://github.com/randohinn/MobileDocRenderer) so I'd have basis on refactoring it a bit.

PR #645, opened 2018-08-31
Fix json format issue in card description

randohinn:
This pull request fixes my issue in #644 by adding the JSON-format required quotation marks around the `src` key in the cards example in [MOBILEDOC.md](https://github.com/bustle/mobiledoc-kit/blob/master/MOBILEDOC.md)


mixonic:
Thanks @randohinn !

Issue #648, opened 2018-10-10
Pasting HTML - wrappers cause unexpected results


ErisDS:
I've searched as best I could but can't figure out if this behaviour is a bug or expected.

This is reproducible for me both by copy-pasting into the [demo editor](https://bustle.github.io/mobiledoc-kit/demo/) and programatically using the DomParser and Mobiledoc Renderer.

Paste: `<p>Hello</p><p>World</p>`

Resulting Mobiledoc.sections:

`
[ [ 1, 'p', [ [ 0, [], 0, 'Hello' ] ] ],
  [ 1, 'p', [ [ 0, [], 0, 'World' ] ] ] ]
`

Paste: `<div><p>Hello</p><p>World</p></div>`

Resulting Mobiledoc.sections:

`[ [ 1, 'p', [ [ 0, [], 0, 'HelloWorld' ] ] ] ]`

I'll try to add a minimal test example to show this in action, but I wanted to raise the issue in advance to check if this is expected behaviour?

If not, I'd be very happy to work towards a fix but might need a little pointer in the right direction. I feel like https://github.com/bustle/mobiledoc-kit/commit/9b4231ebc1fa5740c4f4cc7b768ed55d9cbb6ec9 might be related?

ErisDS:
I'm working on a HTML -> Mobiledoc converter, which is where I'm running into this problem. 

I've setup tests which show this problem with a very minimal reproduction case:

https://github.com/TryGhost/Ghost-SDK/blob/master/packages/html-to-mobiledoc/test/converter.test.js#L24

ErisDS:
This is kind of a long, in depth comment. I'm committed to finding a solution here if I can, although I'm obviously lacking an awful lot of context around the what and why. I would ❤️ any input or feedback!

--------

**Basic problem as I understand it**

I've stepped through what happens in each case with the debugger, and it seems like, with the exception of if there's a Google Docs wrapper, Mobiledoc assumes that what it is processing is always a NodeList that represents sections. As far as I can see, this is a fundamental assumption.

So, in the first instance, SectionParser is called twice, and in the second instance, we only ever enter the SectionParser once, and then recursively iterate over the nodes.

This means that copy-pasting HTML, or trying to convert it as in my case, will always end up in this kind of misinterpretation unless you preprocess it to pass only the sections. 

**In more detail**

It seems to me that issue comes from the following logic:

[...]

PR #651, opened 2018-10-12
walk sections instead of iterating

ErisDS:
closes https://github.com/bustle/mobiledoc-kit/issues/648

I haven't figured out how to write tests for mobiledoc-kit yet. I've done rudimentary testing by writing a test for the 4 cases I outlined in #648 in the html-to-mobiledoc kit lib [here](https://github.com/TryGhost/Ghost-SDK/blob/master/packages/html-to-mobiledoc/test/converter.test.js#L44).

I need to do much more extensive testing around text nodes, and what happens if the DOM is empty, etc etc, but I _think_ this approach is sound.

Will look into writing native tests and wider examples asap.

- Use a dom walker
- Look for valid sections or text nodes
- Parse the sections that are found

### DIFF
--- a/src/js/parsers/dom.js
+++ b/src/js/parsers/dom.js
@@ -16,6 +16,7 @@ import {
   normalizeTagName
 } from '../utils/dom-utils';
 import {
+  contains,
   detect,
   forEach
 } from '../utils/array-utils';
@@ -24,6 +25,7 @@ import { ZWNJ } from 'mobiledoc-kit/renderers/editor-dom';
 
 import SectionParser from 'mobiledoc-kit/parsers/section';
 import Markup from 'mobiledoc-kit/models/markup';
+import { VALID_MARKUP_SECTION_TAGNAMES } from 'mobiledoc-kit/models/markup-section';
 
 const GOOGLE_DOCS_CONTAINER_ID_REGEX = /^docs\-internal\-guid/;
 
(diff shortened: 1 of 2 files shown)


ErisDS:
I've been working through iterations to improve the logic and handle more and more cases, and I'm now down to just a handful:

image

All of these cases seem pretty edge with GSuite's weird structure, but the first one - the DOM parsing of `<b><em>foo</em>bar</b>`, i.e. nested markup with no valid sections, is a real concern.

I think, to solve this case I need to keep some state around, as to whether or not a valid section has been found anywhere in the structure. 

Walking the DOM _feels_ like the correct approach here, even though it's a huge change. However there might be an alternative approach that would involve improving the logic of `detectRootElement` to locate and throw away wrapper elements. Or a hybrid approach combining DOM walking & the original iterations.

I'm a little hesitant to spend too much more time on this without some input, maybe from @bantic on whether you think that DOM walking could ever be a viable approach, or if I've overcomplicated this.

src/js/parsers/dom.js
```diff
@@ -89,6 +101,33 @@ function walkMarkerableNodes(parent, callback) {
...
+  } else if (isValidSection(currentNode)) {
+    callback(currentNode);
+  // Otherwise, look at the next available node
+  } else {
+    if (!currentNode.firstChild) {
+      callback(currentNode);
+    }
+    currentNode = currentNode.firstChild;
+    while (currentNode) {
+      walkSectionNodes(currentNode, callback);
```
mixonic: It would be great to have the recursion flattened out of this function. Instead using a while loop and current node to descend and ascend in a single function. Mostly a style issue, sometimes a performance issue but I doubt so in this case.

ErisDS:
Haven't had chance to review the detailed feedback yet, but just to get the conversation moving on lemme reply to the main point here:

> @ErisDS So I suggest you think of the HTML parse as a "paste" HTML parser. I think your usage of it for legacy document parsing is suitable, but we always presumed that the HTML paste parser would be inappropriate for any number of specific cases.

I experience the same bugs and issues on paste - if you select a large amount of content from E.g. medium and paste it into Ghost's editor, you end up with garbled merged paragraphs and weird styling. 

However, I am definitely trying to kill 2 birds with one stone. If improving the parser means that _both_ pastes and legacy import support are improved, then surely that's just a win-win?

In terms of my intent, medium content is a great example of what I'm trying to get to work - I'll expand out #648 with a non-minimal example. 

Every iteration I've made since my original PR has been to try to get more of the existing test cases to work :)

src/js/parsers/dom.js
ErisDS: Could you expand a little on what you mean here?

ErisDS:
Going to close this PR. Will add an explanation and describe a new approach in the original issue #648.

PR #654, opened 2018-11-19
648 Fix proposal 2: Handle wrappers, siblings and some trees

ErisDS:
This PR is a proposal to fix _most_ of #648, where wrapping, sibling and nested divs cause paragraphs and other "section" elements to be incorrectly merged.

----

The first 4 commits are the same as Proposal 1.

The penultimate commit adds further tests showing additional cases which can be solved with the additional changes.

The last commit moves the recursive search to be performed for each sibling, extending down each branch of the tree until a root / non-wrapper element is found for each branch.

### DIFF
--- a/src/js/parsers/dom.js
+++ b/src/js/parsers/dom.js
@@ -16,6 +16,7 @@ import {
   normalizeTagName
 } from '../utils/dom-utils';
 import {
+  contains,
   detect,
   forEach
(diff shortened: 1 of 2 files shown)


ErisDS:
Updated to remove the "preferred" note - see #648 for why.

ErisDS:
Closing in favour of #658

PR #655, opened 2018-11-19
648 Fix proposal 1: Remove wrappers

ErisDS:
This PR is a proposal to fix _part_ of #648, where wrapping divs cause paragraphs and other "section" elements to be incorrectly merged.

It is **not** the preferred fix. The preferred fix is Proposal 2 😁 

----

The first 3 commits are tests only.

The final commit adds a recursive search for the root element. It attempts to find the innermost wrapper <div> or <section> and use this as the root element. The children of this innermost wrapper are then iterated over as sections, the same as before.

### DIFF
--- a/tests/unit/parsers/dom-test.js
+++ b/tests/unit/parsers/dom-test.js
@@ -69,7 +69,10 @@ let expectations = [
   ['abc\ndef', ['abc def']],
 
   // See https://github.com/bustle/mobiledoc-kit/issues/648
-  ['<div><p>first</p><p>second</p></div>', ['first', 'second']]
+  ['<div><p>first</p><p>second</p></div>', ['first', 'second']],
+  ['<div><div><p>first</p><p>second</p></div></div>', ['first', 'second']],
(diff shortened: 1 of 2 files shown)


tests/unit/parsers/dom-test.js
```diff
@@ -215,6 +233,28 @@ test('strong tag + em + text node creates section', (assert) => {
...
+  let [m1, m2] = markers;
+
+  assert.equal(m1.value, 'stray');
+  assert.equal(m2.value, ' markup tags');
+
+  assert.ok(m1.hasMarkup('b'), 'm1 is b');
+  assert.ok(m1.hasMarkup('em'), 'm1 is em');
+
+  assert.ok(m2.hasMarkup('b'), 'm2 is b');
+  assert.ok(!m2.hasMarkup('em'), 'm1 is not em');
```
kevinansfield:
```suggestion
  assert.ok(!m2.hasMarkup('em'), 'm2 is not em');
```

ErisDS:
Closing in favour of #658

PR #658, opened 2018-11-21
Handle nested section-creating elements in SectionParser

### LINKED ISSUE #648: Pasting HTML - wrappers cause unexpected results

kevinansfield:
closes #654, closes #648 

Updates the `SectionParser` to properly close the current section and start a new one if a "section-creating" element is encountered whilst walking the node tree. This was already being handled for sections added via parser plugins but nested elements were overwriting the current section state rather than creating new sections where necessary.

Also removes the `_parseListElements` function in favour of handling list elements through `parseNode` and `_closeCurrentSection` which allows for extracting nested lists into a single top-level  list section.

Behaviour changes:
- section content with a wrapper element is no longer smushedtogether
- nested lists are treated as a single list (previous behaviour was for all text content of the nested list to be smushedtogether into a single list item)
- list items containing structural sections are parsed as structural sections rather than list items (previous behaviour was for the text content of each section in the `<li>` to be smushedtogether)
  - one exception is list elements containing a single paragraph (`<ul><li><p>content</p></li></ul>`) which will retain the list structure

### DIFF
--- a/src/js/parsers/section.js
+++ b/src/js/parsers/section.js
@@ -150,6 +150,18 @@ class SectionParser {
       let isNestedListSection = isListSection && this.state.section.isListItem;
       let lastSection = this.sections[this.sections.length - 1];
 
+      // we can hit a list item after parsing a nested list, when that happens
+      // and the lists are of different types we need to make sure we switch
+      // the list type back
+      if (isListItem && lastSection && lastSection.isListSection) {
+        let parentElement = node.parentElement;
+        let parentElementTagName = normalizeTagName(parentElement.tagName);
+        if (parentElementTagName !== lastSection.tagName) {
+          this._closeCurrentSection();
+          this._updateStateFromElement(parentElement);
+        }
+      }
+
       // if we've broken out of a list due to nested section-level elements we
       // can hit the next list item without having a list section in the current
       // state. In this instance we find the parent list node and use it to
@@ -159,9 +171,8 @@ class SectionParser {
         !(this.state.section.isListItem || this.state.section.isListSection) &&
         !lastSection.isListSection
       ) {
-        let closestListNode = node.closest(VALID_LIST_SECTION_TAGNAMES.join(','));
         this._closeCurrentSection();
-        this._updateStateFromElement(closestListNode);
+        this._updateStateFromElement(node.parentElement);
       }
 
       // if we have consecutive list sections of different types (ul, ol) then
--- a/src/js/parsers/section.js
+++ b/src/js/parsers/section.js
@@ -80,29 +80,14 @@ class SectionParser {
     if (!finished) {
       let childNodes = isTextNode(element) ? [element] : element.childNodes;
 
-      if (this.state.section.isListSection) {
-        this.parseListItems(childNodes);
-      } else {
-        forEach(childNodes, el => {
-          this.parseNode(el);
-        });
-      }
+      forEach(childNodes, el => {
+        this.parseNode(el);
+      });
     }
 
     this._closeCurrentSection();
 
     return this.sections;
-  }
-
-  parseListItems(childNodes) {
-    let { state } = this;
-    forEach(childNodes, el => {
-      let parsed = new this.constructor(this.builder).parse(el);
-      let li = parsed[0];
-      if (li && li.isListItem) {
-        state.section.items.append(li);
-      }
-    });
   }
 
   runPlugins(node) {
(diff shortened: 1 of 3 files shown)


src/js/parsers/section.js
```diff
@@ -148,6 +154,28 @@ class SectionParser {
       return;
     }
 
+    // handle closing the current section and starting a new one if we hit a
+    // new-section-creating element. Skip if we're currently in a list-item
+    // section so that nested lists can be combined
```
mixonic: @kevinansfield is there a test case for this `Skip if we're currently in a list-item section so that nested lists can be combined`? I don't think so, can it be added?

kevinansfield: @mixonic you're absolutely right, nested grouping wasn't handled at all! I've removed the comment and added a commit with tests of the existing list parsing behaviour to show that it hasn't changed.

I'll push another commit shortly which has a fix for nested lists too but it required some further modification of the section parser. It also has some behaviour changes which _may_ be breaking so I'll comment on those separately.

src/js/parsers/section.js
```diff
@@ -80,31 +80,16 @@ class SectionParser {
...
 
     this._closeCurrentSection();
 
     return this.sections;
   }
 
-  parseListItems(childNodes) {
-    let { state } = this;
-    forEach(childNodes, el => {
-      let parsed = new this.constructor(this.builder).parse(el);
```
kevinansfield: Creating a new SectionParser instance was the main reason for getting rid of this code. To be able to detect nested lists we need access to the current state which is not available in the new instance.

src/js/parsers/section.js
```diff
@@ -155,29 +140,44 @@ class SectionParser {
...
+      let isNestedListSection = contains(VALID_LIST_SECTION_TAGNAMES, tagName)
+        && this.state.section.isListItem;
+
+      if (
+        contains(VALID_MARKUP_SECTION_TAGNAMES, tagName) ||
+        (contains(VALID_LIST_SECTION_TAGNAMES, tagName) && !isNestedListSection) ||
+        contains(VALID_LIST_ITEM_TAGNAMES, tagName)
+      ) {
+        // don't break out of the list for list items that contain a single <p>.
+        // deals with typical case of <li><p>Text</p></li><li><p>Text</p></li>
```
kevinansfield: I added this because I've seen other editors and markdown parsers create this type of structure so I figured it would be good to add specific handling. Otherwise the new behaviour would end up dropping the list altogether, instead creating paragraph sections in place of list sections.

tests/unit/parsers/section-test.js
```diff
@@ -199,12 +199,14 @@ test("#parse handles multiple paragraphs in list item", assert => {
   parser = new SectionParser(builder);
   let sections = parser.parse(element);
 
-  assert.equal(sections.length, 1, "single list section");
+  assert.equal(sections.length, 2, '2 sections');
```
kevinansfield: This is the behaviour change which may need some thought. The previous/current behaviour smushes the text content of all elements in the `<li>` together which isn't ideal but it does preserve the overall "list" structure, the new behaviour gives preference to the nested content which preserves document structure such as headings and paragraphs but removes the list.

src/js/parsers/section.js
```diff
@@ -80,36 +80,27 @@ class SectionParser {
...
-  }
-
   runPlugins(node) {
     let isNodeFinished = false;
     let env = {
       addSection: (section) => {
-        this._closeCurrentSection();
+        // avoid creating empty paragraphs due to wrapper elements around
+        // parser-plugin-handled elements
+        if (this.state.section.isMarkerable && !this.state.text) {
```
kevinansfield: This pattern is repeated a couple of times, I wonder if it would make more sense to move it to `_closeCurrentSection` if we're OK with not copying blank markup sections from the source html?

src/js/parsers/section.js
```diff
@@ -204,7 +238,18 @@ class SectionParser {
       this._createMarker();
     }
 
-    sections.push(state.section);
+    // push listItems onto the listSection or add a new section
+    if (state.section.isListItem && lastSection && lastSection.isListSection) {
+      lastSection.items.append(state.section);
```
kevinansfield: Note to self - add test to ensure consecutive `<ul>`'s do not get merged together.

src/js/parsers/section.js
kevinansfield: > OK with not copying blank markup sections from the source html

To clarify this, the current behaviour with this PR is that blank markup sections are preserved. Eg: `<p>One</p><p></p><p>Two</p>` will result in three markup sections with the middle one being blank.

If we move the duplicated empty-section-removal logic to `_closeCurrentSection` then all blank markup sections would be removed.

kevinansfield: Scratch that - I was apparently mis-remembering behaviour, empty markup sections are ignored in both the `dom` and `section` parser. I'll add tests for such and resolve this

src/js/parsers/section.js
kevinansfield: Added test + fix

kevinansfield:
@mixonic updated to resolve requested changes and improved the behaviour of nested list handling, another 👀 and sanity check on the behaviour would be great when you have a moment 😄

src/js/parsers/section.js
mixonic: @kevinansfield As far as I recall the section parser should only be parsing DOM we create or introduced by `contexteditable` interactions. For example paste events and importing HTML are handled via the DOM parser.

So the implication here is that browsers may insert `p` tags inside lists as the user interacts. Is that what you've seen?

kevinansfield: The DOM parser only grabs top-level elements and relies on the section parser to deal with all the nested elements, in this case the `LI` and `P` are always nested elements as they're inside a `UL`. The changes are primarily to do with handling pastes and using the parsers externally for html->mobiledoc.

mixonic: Yeah, as implemented here:

https://github.com/bustle/mobiledoc-kit/blob/28a746d694e08c7cfc426a4e71a8667b845a1a4d/src/js/parsers/dom.js#L140

You are correct. I feel like I should have caught this earlier, but if these are explicitly corrections to the paste parsing then we should implement them as corrections in the DOM parser before it passes content to the section parser. The section parser (in theory) should only be concerned with contexteditable-type inputs, not with any plausible HTML input from a paste. The latter is basically an unbounded set of concerns, and it is nice to keep them out of the editor parser.

tests/unit/parsers/section-test.js
mixonic: We discussed this on Slack. When the DOM inside a `<li>` is trivial as in this test example you might want the smushing behavior, but as soon as there is arbitrarily complex content in the list item (many paragraphs, images, etc) considering this content as a candidate for top level sections is better behavior.

mixonic:
I'm a go on these changes. @kevinansfield has a test failure he just discovered as we went over this, so we should wait on his work to proceed with merge.

kevinansfield:
> @kevinansfield has a test failure he just discovered as we went over this, so we should wait on his work to proceed with merge.

Additional tests for the discovered error and associated fixes have been added 😄

sdhull:
Thank you @kevinansfield!!  🎉 🎉

Issue #548, opened 2017-04-12
Entering logograms issues in macOS


disordinary:
This is the issue from the Ghost project.
https://github.com/TryGhost/Ghost/issues/8319

(The issue was submitted by @mach)

# Issue Summary

On Ghost Alpha's new text editor, Japanese text can't be typed correctly with Japanese IME (input method).

![image](https://cloud.githubusercontent.com/assets/82702/24944278/6743d564-1f95-11e7-80c8-eeedab8bbb7a.gif)

Steps to Reproduce

    Select Japanese IME (either Apple's JapaneseIM or Google's Japanese IME)
    Type "watasi"
    Press Enter key

Expected Result

At step 2: "watasi" should be converted to "わたし"
At step 3: "わたし" should be converted to "私"
Actual Result

At step 2: "watasi" is converted to "wあたし"
At step 3: "wあたし" is converted to "w"
Technical details:

    Ghost Version: 1.0.0-alpha.19
    Node Version: v6.10.2
    Browser/OS: Google Chrome / macOS Sierra
    Database: MySQL 5.7.17

disordinary:
From testing it seems that it only occurs in an empty paragraph.

taking:
Korean has the same problem.

Adrift001:
Isn't this bug going to be fixed?

kwmr:
When will they fix it?
I'm tired of waiting.

taking:
> When will they fix it?
> I'm tired of waiting.

Haha not fix.

sxueck:
It’s still half a year before 2020. But still not fixed, who knows what other editors?

imyhui:
It’s been two years, still this error

cmsax:
@soulteary addressed the problem, ref: [Issue 9801](https://github.com/TryGhost/Ghost/issues/9801).
However it's still not merged. The bug still there.

kwmr:
uuuum...
The bug still there.

cmsax:
Insert `markdown` block as a workaround.

Smith-Cruise:
we are still enjoying this bug :)

wbpmrck:
we are still enjoying this bug :)

JianpingZong:
what's the essential thing of human being?
we are still enjoying this bug :)

bin16:
The bug has been in ghost for two years, now 3.0 released...

bin16:
[...]

PR #661, opened 2018-12-13
WIP ignore IME key clicks

iezer:
Partial fix for #548. This fixes Korean and Japanese input (using Input Method Editor IME) for most cases. 

Editor is still broken when entering text at the beginning of a section. Still working on that. Fixing this will probably involve fixing #589, ie. don't do a full re-render when not needed. This also effects spellcheck and Android input.

Basically the keys Enter, Backspace, Space, Tab, and Arrows are all used to navigate and select within the IME editor so we want mobiledoc to ignore those.

Illustration of the bug:

**Pressing Enter**

**Pressing Backspace**
 * Pressing backspace seems to duplicate the previous character

**After fixed**

**Still broken at start of section**
 * Notice the "w" goes straight to the document and is not part of the IME input

### DIFF
--- a/src/js/editor/event-manager.js
+++ b/src/js/editor/event-manager.js
@@ -166,6 +166,11 @@ export default class EventManager {
     let range = editor.range;
 
     switch(true) {
+      case key.isIME(): {
+        // For IME（eg. typing in Japanese or Korean), we want to ignore all keydowns,
+        // especially ENTER, DELETE, and TAB, which all have special behaviors in the IME popover.
+        break;
+      }
       // FIXME This should be restricted to only card/atom boundaries
       case key.isHorizontalArrowWithoutModifiersOtherThanShift(): {
         let newRange;
--- a/tests/acceptance/basic-editor-test.js
+++ b/tests/acceptance/basic-editor-test.js
@@ -183,3 +183,18 @@ test('prevent handling newline', (assert) => {
   Helpers.dom.insertText(editor, ENTER);
   assert.postIsSimilar(editor.post, expected);
 });
+
+// These keys have meaning with the IME window so should be ignored.
+const IGNORED_IME_KEYS = [ 'Enter', 'Tab', 'ArrowLeft', 'ArrowRight', ' ', 'Backspace'];
+
+IGNORED_IME_KEYS.forEach(function(key) {
+  test(`pressing "${key}" in Input Method Editor (IME) is ignored`, (assert) => {
+    editor = Helpers.editor.buildFromText('hi|hey', {element: editorElement});
+
+    assert.hasElement('#editor p:contains(hihey)');
+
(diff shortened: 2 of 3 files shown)


mixonic:
@iezer are there docs for the IME APIs? I was looking into it and I wasn't sure if FireFox and Chrome have the same APIs here, or if they are standard/non-standard etc.

Looking good.

iezer:
@mixonic I found some docs here but I'm not sure if they're relevant. https://www.chromium.org/developers/design-documents/extensions/proposed-changes/apis-under-development/input-method-editor

The IME interface is at the OS level so it's the same on all browsers and text-editors, even in the terminal!

I just tested this on Firefox and Safari and it works.

Ferrari:
Sorry for asking, any update on this PR? Hope this could merge soon and the IME issue could be solved. :)

kwmr:
If this bug remains for a long time, no Japanese or Koreans will use it. I WANT SOME- THING TO FIX IT...

hq5544:
This problem troubles me such a long time.
Hope this PR can be merged soon.

tohutohu:
I strongly hope that this PR will be merged early.
If there is a reason for not being merged, please tell us why.

xufangda:
Merge it Please!!!

bin16:
@gpoitch The PR is important for all users who need input method.

gpoitch:
@bin16 the PR is still incomplete according to the author @iezer.  Happy to help push it along if the author or someone else can step up and finish it or confirm it's working as-is.  I am no longer contributing to this project.

xufangda:
> @bin16 the PR is still incomplete according to the author @iezer. Happy to help push it along if the author or someone else can step up and finish it or confirm it's working as-is. I am no longer contributing to this project.


Thanks so much for reply. This PR has been proved to be effective by @soulteary.  The author @iezer has pointed out that this PR solves problem #548, and this is the biggest problem for every asian user.

The so called incomplete refers to issue #589. We can wait for people proposing a new PR. Please merge this first.

bin16:
@gpoitch It looks good at least for this issue, based on screenshot of the comment(https://github.com/bustle/mobiledoc-kit/pull/661#issuecomment-562331198) by @xufangda 

> Editor is still broken when entering text at the beginning of a section. 

include this?

bin16:
I've checked, the PR is a partial fix. It's broken when you insert with IME to an empty section. e.g. typing `ceshi` with IME, `c` is caught by editor, and IME take `eshi` only.

bin16:
@iezer You can replace `key.isIME()` with `event.isComposing`, https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing

The API is not supported by some browsers(IE), but it would not be worse. https://caniuse.com/#feat=mdn-api_keyboardevent_iscomposing.

ZeeJab:
Superseded by #738

Issue #570, opened 2017-08-10
Improve build tooling and package consumption


gpoitch:
For example, to access the parser in node environment, you currently have to do this:

```js
const parser = require('mobiledoc-kit/dist/commonjs/mobiledoc-kit/parsers/mobiledoc').default
```

instead of:
```js
const parser = require('mobiledoc-kit/parsers/mobiledoc')
```

It's also not providing entry points for module bundlers that can take advantage of tree-shaking.
I can lend a hand by using rollup instead of broccoli.

iezer:
@gpoitch Thanks for opening this issue. I've worked on modernizing the build by using rollup and I have a branch here. https://github.com/bustle/mobiledoc-kit/tree/iezer/rollup It allows to to remove the outdated dependencies on `broccoli-multi-builder` and `broccoli-test-builder`.

I'm not sure this solves your use-case because rollup by definition only exposes functionality that is exported in `src/js/index.js`. That does not currently include the parser. Do you think the parser should be part of the object exported from `index.js`? Or perhaps exported as a separate AMD module?

I'd appreciate it if you could take a look at my branch, try out the build locally and let me know if it solves your use case. I'd also appreciate your feedback on if there's anyway rollup should be configured differently.

So far I'm noticing a very slight decrease in payload size with Rollup (The un-minified AMD build went down 7KB from 544 KB to 537 KB) but build times are slightly higher.

CC @rondale-sc

PR #662, opened 2018-12-21
Export MOBILEDOC_VERSION from index

iezer:
As part of modernizing the build using [Rollup.js](https://rollupjs.org/guide/en), we want Mobiledoc users to get away from importing from any module other than `mobiledoc-kit`. The one use case I've seen for other packages is that many apps do something like this:

```js
import { MOBILEDOC_VERSION } from 'mobiledoc-kit/renderers/mobiledoc';  
const EMPTY_MOBILEDOC = {
  version: MOBILEDOC_VERSION,
  //...
}`
```

Examples: [ember-yapp-mobiledoc-editor](https://github.com/lukemelia/ember-yapp-mobiledoc-editor/blob/4dddb572032ae216a2df23269b2dd4eba09d7a16/tests/dummy/app/controllers/index.js#L2), [queertangoclub](https://github.com/tim-evans/queertangoclub/blob/32564b8c4981f0112f4e2de7c52fea805ac09c81/app/components/text-editor/component.js#L4), [Ghost Admin](https://github.com/TryGhost/Ghost-Admin/blob/84059cd5a7451c5fa49f78f5209953f6bdbe5fca/lib/koenig-editor/addon/components/koenig-basic-html-input.js#L2-L7), [ember-mobiledoc-editor](https://github.com/bustle/ember-mobiledoc-editor/search?q=MOBILEDOC_VERSION&unscoped_q=MOBILEDOC_VERSION)

So to move away from there I'd like to add `MOBILEDOC_VERSION` as a named export to the `mobiledoc-kit` module so that those import states can change to `import { MOBILEDOC_VERSION } from 'mobiledoc-kit';`.

Once this PR is merged, we can open PR against those other repos.

It may also be useful to export the default `mobiledocRenderers` function but I haven't found any consumers of that yet. If we need it we can export in a separate PR.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -5,6 +5,7 @@ import Range from './utils/cursor/range';
 import Position from './utils/cursor/position';
 import Error from './utils/mobiledoc-error';
 import VERSION from './version';
+import { MOBILEDOC_VERSION } from 'mobiledoc-kit/renderers/mobiledoc';
 
 const Mobiledoc = {
   Editor,
@@ -13,12 +14,13 @@ const Mobiledoc = {
   Range,
   Position,
   Error,
-  VERSION
+  VERSION,
+  MOBILEDOC_VERSION
 };
 
 export function registerGlobal(global) {
   global.Mobiledoc = Mobiledoc;
 }
 
-export { Editor, UI, Range, Position };
+export { Editor, UI, Range, Position, MOBILEDOC_VERSION };
 export default Mobiledoc;


iezer:
Based on the use-case described in #570 it might make sense to also export the `mobiledocParsers` function from `/parsers/mobiledoc/index`

iezer:
@mixonic It would be great to merge and release this non-breaking change. We can then start updating mobiledoc-kit consumers as part of the process to migrate the build to rollup (#570).

CC @rondale-sc

mixonic:
Merging this as a first change to the top-level exports. We can add others as needed.

PR #664, opened 2019-01-16
[WIP] Use Rollup to package mobiledoc-kit

rondale-sc:
This is the main branch to update the build system of mobiledoc-kit to use rollup instead of the https://github.com/bustle/broccoli-multi-builder.

We now have rollup producing a `cjs`, `amd`, `global`, and `esm` package in the final distribution.  The test file similarly uses rollup to wrap up all of the tests and associated dependencies.

Work remaining:

- [x] Audit packages that depend on old AMD packages e.g. `import X from ‘mobiledoc-kit/something` to find all the non-top-level imports.  Rollup only produces a single `define` in AMD as it "roll"s up everything inside of that.
- [ ] Fix the circular dependency between cursor/position and cursor/range
- [ ] Fix the circular dependency between parsers/dom and parsers/section
- [x] Release PR https://github.com/bustle/mobiledoc-kit/pull/662 and audit associated addons that depend on `MOBILEDOC_VERSION`
- [x] Receive general feedback from consumers

List of known consumers from libraries.io:

- Bustle/romper
- yapp
- cardstack
- ghost
- ember-template-input
   - Already tested a version of this package with this addon and tests were green

To test this I'll be using `npm pack` on this branch to produce a tar file that is representative of how NPM itself would serve the package and install it in our audit addon/library/repository and run their tests to ensure no (or at least limited) breakage.

/cc @iezer


kevinansfield:
Regarding Ghost's usage of mobiledoc-kit, we have two repos [Ghost](https://github.com/tryghost/ghost) and [Ghost-Admin](https://github.com/TryGhost/Ghost-Admin) although both are currently using a fork. I'm happy to help with the audit of those 🙂

rondale-sc:
@kevinansfield 	That'd be very helpful, if you'd like an extra set of eyes I'm available to help with any snags you encounter.

kevinansfield:
Regarding non-top-level imports we currently have these across our projects using mobiledoc-kit:

[Ember application](https://github.com/TryGhost/Ghost-Admin):
- `import {getContentFromPasteEvent} from 'mobiledoc-kit/utils/parse-utils';`
- `import {getEventTargetMatchingTag} from 'mobiledoc-kit/utils/element-utils';`
- `import {MOBILEDOC_VERSION} from 'mobiledoc-kit/renderers/mobiledoc';`
- `import Browser from 'mobiledoc-kit/utils/browser';`
- `import Editor from 'mobiledoc-kit/editor/editor';`
- `import Key from 'mobiledoc-kit/utils/key';`
- `import MobiledocRange from 'mobiledoc-kit/utils/cursor/range';`

[Node html-to-mobiledoc library](https://github.com/TryGhost/Ghost-SDK/tree/master/packages/html-to-mobiledoc):
- `const DOMParser = require('mobiledoc-kit/dist/commonjs/mobiledoc-kit/parsers/dom').default;`
- `const Builder = require('mobiledoc-kit/dist/commonjs/mobiledoc-kit/models/post-node-builder').default;`
- `const mobiledocRenderer = require('mobiledoc-kit/dist/commonjs/mobiledoc-kit/renderers/mobiledoc').default;`

rondale-sc:
@kevinansfield What distribution does the Ember application use?  I'm assuming the AMD version?

rondale-sc:
I think we could expose these in the main entry point such that you could for instance:

```js
import Mobiledoc from 'mobiledoc-kit';

const { getContentFromPasteEvent } = Mobiledoc;
```

kevinansfield:
Yes, we're pulling in the AMD version. Being able to pull those off the main entry point would be good 👍

ErisDS:
Ghost also depends on Mobiledoc-Kit in our HTML -> Mobiledoc conversion tool which can be found here: https://github.com/TryGhost/Ghost-SDK/blob/master/packages/html-to-mobiledoc/lib/converter.js

You can see that it only depends on a small-subsection of mobiledoc-kit + JSDOM, but I have to require the whole thing.

Hopefully the rollup build will solve the awkward requiring here. It would be awesome to see mobiledoc-kit also broken into smaller pieces (hopefully easier after decoupling circular dependencies) so we don't have to require the whole kit for this parsing exercise.

rondale-sc:
@ErisDS I think that providing a module entry point should enable you to require with the actual modules themselves and you could then rollup to "treeshake" the unused stuff yourself.  I'm still learning how the module system works with node though, I'm currently thinking that I'll need to package .mjs files in order for that to work. For now we'll continue shipping the commonjs, but we'll make the things necessary for html-to-mobiledoc exported on the main entry point for mobiledockit (that's on my todolist today)

rondale-sc:
@kevinansfield I believe the modules are now exported in both the default export object and named exports.  Should be good to go for CJS / AMD now, can you take a quick look at https://github.com/bustle/mobiledoc-kit/pull/664/commits/a4d6a2ba8b69a366cbf2c75a8aebdf7b8d903b6e and verify that is what you need?

ZeeJab:
I'm not sure if using Rollup for the AMD build is the right approach if it causes large breaking changes to the public module API.

Is the motivation for this change mostly to modernize the build pipeline and get off the `broccoli-multi-builder` dependency? I think it would be better to use something like Babel to do the JS module to AMD transpilation since it can preserve module names and not break the public API.

At Bustle, other than our Ember app where we're using the AMD build, we're also using [react-mobiledoc-editor](https://github.com/joshfrench/react-mobiledoc-editor) which is built with webpack. I just took a look and it looks like we're getting the CommonJS build in that case. Moving to named exports from the package root for the JS module build seems fine, although it would technically be possible to support the same nested modules (e.g. `mobiledoc-kit/utils/cursor/range`) if we wanted to. But because these weren't well supported in existing webpack builds I don't think not supporting it will cause API breakage.

ZeeJab:
I believe this is where the AMD transpilation is happening with Babel in the Ember build for reference: https://github.com/emberjs/ember.js/blob/master/broccoli/to-named-amd.js

rondale-sc:
@ZeeJab Yeah, the idea of using Rollup was indeed primarily to modernize the build pipeline and to get rid of the multipart builder with the hope that it'd shrink the package size down a smidge.  

Within mobiledoc-kit itself the modules can still be imported as before.  Because of this I'm not sure if this represents a change to the public api so much as it forces us to define what the public API is more concretely where before the consuming loader had access to all modules.  Though I'm definitely still learning the different module bundling strategy's pros/cons.

I did try to use the build with react-mobiledoc-editor (RME) / webpack with this rollup build last week and got to only 2 failing tests. Since we provide a module entry point in the package.json I wonder if RME could import the modules directly and have webpack sort it all. https://github.com/rollup/rollup#publishing-es-modules.

Similarly, for Ghost as mentioned above the actual surface of the API that is used there is minimal.

If we decide we'd like to ensure modules are preserved as they are now perhaps we can update multipart builder to use a newer version of babel and see what that looks like.

ErisDS:
> Similarly, for Ghost as mentioned above the actual surface of the API that is used there is minimal.

Just to be clear, deep requires in node land are possible, but it's fairly obviously "Private API". Therefore, I expect it to break at some point and that I'd have to update those deep requires. The _hope_ is that with this PR it'll be replaced with Public API that I can depend on 😄

ZeeJab:
Closing. This PR is now preceded by https://github.com/bustle/mobiledoc-kit/pull/710

PR #665, opened 2019-01-16
Add deprecation around DOM / Text Renderer removal from mobiledoc-kit

rondale-sc:
In the next major bump I believe we can remove our dependency on
Domrenderer and TextRenderer here with little downsides.  The standalone
repositories would continue to exist as they do today this function
would be only for the `mobiledoc` format which will return a "mobiledoc"
that can be passed into the standalone Renderers to reterieve their
output.

The main advantage this is that it decouples us from these renderers.

The work that'd need to happen to land this is as follows:

- [ ] remove the `format` switch and update the assertion to throw when
anything other than "mobiledoc" is passed to this function
- [ ] Update the documentation to show examples of how to do this
properly with the standalone renderers

### DIFF
--- a/src/js/editor/editor.js
+++ b/src/js/editor/editor.js
@@ -24,6 +24,7 @@ import {
 } from './key-commands';
 import { CARD_MODES } from '../models/card';
 import assert from '../utils/assert';
+import deprecate from '../utils/deprecate';
 import MutationHandler from 'mobiledoc-kit/editor/mutation-handler';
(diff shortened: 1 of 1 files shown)


rondale-sc:
Was looking to remove this in another PR and cross reference.  But I've found that we use these renderers to enable copy/paste functionality.  

I'm looking into a way to remove these without breaking this functionality, and will update this PR soon to reflect my findings.

Issue #672, opened 2019-02-14
Parsing indented HTML throws errors and creates unnecessary whitespace


kevinansfield:
When DOM is created from HTML that has been indented it contains many extra blank (but lengthy) `textNode`s. This was causing two problems:

1. The `SectionParser` throws errors because `textNode`s can appear inside non-markerable sections (see error screenshot below)
2. Sections can frequently end up with large amounts of whitespace at the beginning and end of the content. For rendering this is not much of a problem because browsers ignore insignificant whitespace, however the primary use-case of the DOM/Section parsers is to create mobiledoc for display in the editor - in this case all whitespace can be significant.

Error when parsing content such as:

```html
<ul>
    <li>
        One
    </li>
</ul>
```

<img width="794" alt="screenshot 2019-02-14 at 13 42 27" src="https://user-images.githubusercontent.com/415/52797758-6cb75200-306e-11e9-8e5a-ea6bc4013bee.png">

PR #673, opened 2019-02-14
Fix error and extra leading/trailing whitespace when parsing indented HTML

kevinansfield:
closes https://github.com/bustle/mobiledoc-kit/issues/672
- do not create markers for non-markerable sections when closing sections in the SectionParser
- add a utility method to trim whitespace from the beginning/end of a section
- trim whitespace from sections after parsing list items in the SectionParser
- trim whitespace from all top-level parsed sections in the DOMParser

### DIFF
--- a/src/js/parsers/section.js
+++ b/src/js/parsers/section.js
@@ -238,7 +238,7 @@ class SectionParser {
     let { state } = this;
 
     const markups = this._markupsFromElement(element);
-    if (markups.length && state.text.length) {
+    if (markups.length && state.text.length && state.section.isMarkerable) {
       this._createMarker();
     }
     state.markups.push(...markups);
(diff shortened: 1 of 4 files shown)


src/js/parsers/section.js
```diff
@@ -341,9 +351,11 @@ class SectionParser {
 
   _createMarker() {
     let { state } = this;
-    let text = transformHTMLText(state.text);
-    let marker = this.builder.createMarker(text, state.markups);
-    state.section.markers.append(marker);
+    if (state.section.isMarkerable) {
```
mixonic: Should this check be in `parseElementNode`? https://github.com/bustle/mobiledoc-kit/blob/6f57a62f5979339a9f5a6e273bb5d58d1cc4681d/src/js/parsers/section.js#L238 It seems less than ideal that `_createMarker` has an internal guard to sometimes *not* do what it is presumed to do. I think we should be catching this a bit higher up.

kevinansfield: 👍 updated to move checks to where we're calling `_createMarker`

Issue #676, opened 2019-04-29
range#expandByMarker will not expand to the beginning of the section


kevinansfield:
If you have a range that you want to expand to include the full range of a markup, and that range extends to the beginning of the section, `#expandByMarker` will stop at the end of the range's head marker rather than all markers that match the detect callback.

The problem is occurring [in this code](https://github.com/bustle/mobiledoc-kit/blob/master/src/js/utils/cursor/range.js#L140-L146) because we look for "firstNotMatchingDetect" however we won't reach a non-matching detect because we reach the beginning of the section first. When there's no non-matching detect we revert to the default behaviour of using the marker that's under the range's head rather than the beginning of the section.

I ran into this [issue in Ghost](https://github.com/TryGhost/Ghost/issues/10717), here's a screenshot to help illustrate:

![](https://user-images.githubusercontent.com/415/56890981-456ffe80-6a7b-11e9-8d67-9036493f9b54.gif)

PR #677, opened 2019-04-29
🐛 Fixed range#expandByMarker not expanding to beginning/end of section

kevinansfield:
closes https://github.com/bustle/mobiledoc-kit/issues/676
- after failing to find a non-matching marker, perform a check for the first/last markers in the range's section matching the supplied detection callback. If it does match we can assume that we checked all markers in that direction so we expand to the beginning/end of the section
- improves behaviour to match the documentation: "expand a range to all markers matching a given check"

### DIFF
--- a/src/js/utils/cursor/range.js
+++ b/src/js/utils/cursor/range.js
@@ -141,12 +141,20 @@ class Range {
       return !detectMarker(i);
     };
 
-    let headMarker = head.section.markers.detect(firstNotMatchingDetect, head.marker, true);
-    headMarker = (headMarker && headMarker.next) || head.marker;
+    let headMarker = headSection.markers.detect(firstNotMatchingDetect, head.marker, true);
+    if (!headMarker && detectMarker(headSection.markers.head)) {
+      headMarker = headSection.markers.head;
+    } else {
+      headMarker = (headMarker && headMarker.next) || head.marker;
+    }
     let headPosition = new Position(headSection, headSection.offsetOfMarker(headMarker));
 
     let tailMarker = tail.section.markers.detect(firstNotMatchingDetect, tail.marker);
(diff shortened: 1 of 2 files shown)


src/js/utils/cursor/range.js
```diff
@@ -141,12 +141,20 @@ class Range {
       return !detectMarker(i);
     };
 
-    let headMarker = head.section.markers.detect(firstNotMatchingDetect, head.marker, true);
-    headMarker = (headMarker && headMarker.next) || head.marker;
+    let headMarker = headSection.markers.detect(firstNotMatchingDetect, head.marker, true);
+    if (!headMarker && detectMarker(headSection.markers.head)) {
+      headMarker = headSection.markers.head;
+    } else {
+      headMarker = (headMarker && headMarker.next) || head.marker;
```
mixonic:
```suggestion
      headMarker = headMarker.next || head.marker;
```

src/js/utils/cursor/range.js
```diff
@@ -141,12 +141,20 @@ class Range {
...
+    }
     let headPosition = new Position(headSection, headSection.offsetOfMarker(headMarker));
 
     let tailMarker = tail.section.markers.detect(firstNotMatchingDetect, tail.marker);
-    tailMarker = (tailMarker && tailMarker.prev) || tail.marker;
+    if (!tailMarker && detectMarker(headSection.markers.tail)) {
+      tailMarker = headSection.markers.tail;
+    } else {
+      tailMarker = (tailMarker && tailMarker.prev) || tail.marker;
```
mixonic:
```suggestion
      tailMarker = tailMarker.prev || tail.marker;
```

kevinansfield:
@mixonic requested changes reviewed and committed 👍

PR #670, opened 2019-01-29
WIP - Initial pass to get attributes in the right location

rondale-sc:

PR #681, opened 2019-05-07
Bump spec to 0.3.2, optional section attributes

mixonic:
Bump the Mobiledoc spec to 0.3.2.

This adds an optional "section attributes" item to markup sections and list sections.

* The attributes are encoded as a list of key/value pairs, for example: `['data-md-text-align', 'center']`
* Only `data-md-text-align` is permitted by the spec. An alternative would be to permit arbitrary attribute values, but claim the `data-md-` namespace and only restrict values added there.
* The attributes could alternatively been located in a new top-level property of the spec, but I chose to simply add them to each section instead. It seems fine for an initial pass, if every section will have an attribute we might want to encode them in a different way.
* I've added this to markup and list sections. Alternatives would be a) just markup sections b) markup sections and list *item* sections.

Any thoughts? I've got another PR incoming that polishes @rondale-sc's prior work to implement this in Mobiledoc-kit.

This is an extraction/edit from WIP https://github.com/bustle/mobiledoc-kit/pull/670

/cc @ZeeJab @kevinansfield @rondale-sc

### DIFF
--- a/MOBILEDOC.md
+++ b/MOBILEDOC.md
@@ -36,7 +36,7 @@ The wrapper signature:
 
 ```
 {
-  version: "0.3.1",                         ──── Versioning information
+  version: "0.3.2",                         ──── Versioning information
   markups: [                              ──── Ordered list of markup types
     markup,
     markup
@@ -65,7 +65,7 @@ Markups have a tagName and an optional array of attributes. Not all markups can 
 
 ```
 {
-  version: "0.3.1",
+  version: "0.3.2",
   markups: [
     [tagName, optionalAttributesArray],   ──── Markup
     ["em"],                               ──── Example simple markup with no attributes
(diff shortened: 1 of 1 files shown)


YoranBrondsema:
I am continuing work on this as we'd like to offer text alignment in Sutori.

> Only data-md-text-align is permitted by the spec. An alternative would be to permit arbitrary attribute values, but claim the data-md- namespace and only restrict values added there.
I'd be in favor of only permitting `data-md-text-align` attributes for now, and then see what use cases come up. Then eventually extending it or permitting any attribute. I agree that all attributes set by Mobiledoc should be prefixed with `data-md-`.

> I've added this to markup and list sections. Alternatives would be a) just markup sections b) markup sections and list item sections.
I tried all three and from a UX perspective, I think you picked the right one.

mixonic:
@YoranBrondsema I think the only real blocker on this is the renderers themselves needing to support the spec. Want to schedule a pairing session to try and wrap it up? I'll send an email.

PR #687, opened 2019-07-15
Remove node_modules/.bin path from scripts in package.json

YoranBrondsema:


YoranBrondsema:
Looks like something is wrong with the tests running on Sauce Labs.

mixonic:
I tested this locally and it seems correct. Perhaps we had those hacks and they were needed on older versions of npm. I'm not totally sure.

PR #690, opened 2019-07-16
Fix #689: Original range should be remembered after `toggleSection`

### LINKED ISSUE #689: Original range should be remembered after `toggleSection`
Right now, after toggling a section, the cursor is reset to the beginning of the section and the selection is cleared:

mobiledoc-kit-remember-range

Ideally, it would be like on Quill, where the range is remembered:

quill-remember-range

YoranBrondsema:
### Fix #689
The idea is that when toggling a section, we maintain a list of mappings from the original to the new section. Sometimes, sections are maintained but the tag name is simply modified. This happens for instance when going from a paragraph to a H1. However, sometimes sections are destroyed and a new one is created. For instance, when going from a list to a paragraph.

By looking at the mapping of old section to new section, we can reconstruct the selection after toggling:

mobiledoc-kit

The only edge case is when toggling to a list. There, the sections in the range have to point to ListItem's rather than ListSections. But at the time of toggling, we only have access to the ListSection, and we don't know which ListItem's the new range should point to.

mobiledoc-kit

### Other changes
1. Remove redundant CSS from demo.css
2. Add a button in the demo to toggle a quote

### DIFF
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -790,7 +790,6 @@ class PostEditor {
 
     sectionTagName = normalizeTagName(sectionTagName);
     let { post } = this.editor;
-    let nextRange = range;
 
     let everySectionHasTagName = true;
     post.walkMarkerableSections(range, section => {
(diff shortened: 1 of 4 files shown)


src/js/editor/post.js
```diff
@@ -800,18 +799,47 @@ class PostEditor {
...
-    }
+    let nextRange = this._determineNextRangeAfterToggleSection(range, sectionTransformations);
     this.setRange(nextRange);
   }
 
+  _determineNextRangeAfterToggleSection(range, sectionTransformations) {
+    if (sectionTransformations.length) {
+      let changedHeadSection = sectionTransformations
+        .find(({ from }) => { return from === range.headSection; })
+        .to;
```
mixonic: So this was described to me as: Walk through the changes, find the mapping which is for the same section as the old range's head, then use the mapped "to" as the new section for the head.

This seems good and like it might be a useful pattern in refactoring the range code in general.

YoranBrondsema:
@mixonic Just realized that we didn't merge this and didn't include it in the pre-release. Are you cool merging this and then publishing a new pre-release?

PR #692, opened 2019-07-17
Set default range for 'setAttribute' in PostEditor

YoranBrondsema:
Otherwise the range must be specified every time this public API is used., i.e. requires something like

`postEditor.setAttribute('text-align', 'center', `postEditor._range);`

instead of

`postEditor.setAttribute('text-align', 'center');`


YoranBrondsema:
Only Sauce Labs tests are failing.

PR #710, opened 2020-02-05
Use Rollup 🍣 for build

ZeeJab:
_Previously on Love Island_...

This PR supercedes https://github.com/bustle/mobiledoc-kit/pull/664

This is a **major** release. The motivation behind this PR is to modernize and simplify Mobiledoc Kit. I've added rollup to build the tests and the library code for distribution which allows us to remove dependence on Broccoli 🥦 (some of the broccoli libraries used in Mobiledoc Kit are no longer maintained which was making working on the code quite difficult.)

Going forward we will also only publish this library in ESM format, and no longer support AMD, CommonJS or globals builds. 

This PR also removes IE 11 tests.

Still left to do:

- [x] Building the demo app
- [x] Update website deployment
- [x] Remove globals support
- [x] Update README
- [x] Pass CI tests
- [ ] Verify against real world uses


ZeeJab:
@kevinansfield would you mind looking through this PR to see if I overlooked anything that may have been important to ghost? Otherwise I will do a pre-release so we can test out in our apps.

Thanks!

RuslanZavacky:
Hi @ZeeJab 

We are using something similar to what Ghost is using in https://github.com/TryGhost/SDK/blob/main/packages/html-to-mobiledoc/lib/converter.js#L1 and it seems that currently with rollup, none of those are exported:

DOMParser, PostNodeBuilder and mobiledocRenderers

I am not familiar with rollup, but is it possible to add them to be re-exported? That would simplify some work on top of mobiledoc-kit for rendering mobiledoc -> html -> mobiledoc.

Thank you!

PR #712, opened 2020-02-19
Bug Fix: don't modify tail.section if offset is 0 (aka tail.section is not selected)

nikse:
Fixes 2 bugs around double clicking a paragraph to modify/copy-paste it:

1) Issue: if a user double clicks to select and copy a paragraph that has a card following it, the card gets caught up in the copy and will also be pasted wherever you try to paste the paragraph
Fix: rather than copying the card we now insert a new blank paragraph

2) Issue: if a user double clicks to select a paragraph to add section level markup and there is another paragraph following the selected one the second paragraph will also have the section level markup applied to it
Fix: only apply markdown to selected sections

### DIFF
--- a/assets/demo/debug.html
+++ b/assets/demo/debug.html
@@ -1,70 +1,81 @@
 <!DOCTYPE html>
 <html>
-<head>
-  <meta charset="utf-8">
-  <title>Mobiledoc Kit Debug</title>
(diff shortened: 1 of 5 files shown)


nikse:
closing due to editor auto linting. will reopen in a new clean pr

PR #713, opened 2020-02-19
[Bug Fix] don't modify tail.section if offset is 0 (aka tail.section is not selected)

nikse:
Fixes 2 bugs around double clicking a paragraph to modify/copy-paste it:

1. Issue: if a user double clicks to select and copy a paragraph that has a card following it, the card gets caught up in the copy and will also be pasted wherever you try to paste the paragraph
Fix: rather than copying the card we now insert a new blank paragraph

2. Issue: if a user double clicks to select a paragraph to add section level markup and there is another paragraph following the selected one the second paragraph will also have the section level markup applied to it
Fix: only apply markdown to selected sections

### DIFF
--- a/src/js/editor/post.js
+++ b/src/js/editor/post.js
@@ -15,6 +15,16 @@ function isListSectionTagName(tagName) {
   return tagName === 'ul' || tagName === 'ol';
 }
 
+function shrinkRange(range) {
+  const { head, tail } = range;
+
+  if (tail.offset === 0 && head.section !== tail.section) {
+    range.tail = new Position(tail.section.prev, tail.section.prev.length);
+  }
+
+  return range;
+}
+
 const CALLBACK_QUEUES = {
   BEFORE_COMPLETE: 'beforeComplete',
   COMPLETE: 'complete',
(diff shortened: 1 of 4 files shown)


ZeeJab:
@nikse Made a small change to your PR. How does this look?

1. I don't think hardcoding special behavior for `ul` is necessary. Why does the special behavior apply to `ul` but not `ol`? It looks like this was here just to satisfy some tests that happened to use `ul`. Is there a good reason for this special casing?
2. Rather than pushing logic for the special case throughout the `toggleMarkup` function, in this commit we can normalize the range object at the beginning of the method to only include the sections we want to be modified.

nikse:
> 1. I don't think hardcoding special behavior for `ul` is necessary. Why does the special behavior apply to `ul` but not `ol`? It looks like this was here just to satisfy some tests that happened to use `ul`. Is there a good reason for this special casing?
- I think this was because ul's were already acting correctly, i.e. adding a blank bullet at the end of the list in this situation

> 2. Rather than pushing logic for the special case throughout the `toggleMarkup` function, in this commit we can normalize the range object at the beginning of the method to only include the sections we want to be modified.
- I think this sounds good, I'll test it out again (I'm a bit rusty on what I did here)

ZeeJab:
@nikse why would `ul`s behave differently than `ol`s? what were the steps to reproduce so I can test out? I just want to make sure there's not something I'm missing. Otherwise the changes I made to this branch should be good and we can merge 👍

PR #733, opened 2020-05-29
Typescriptify Phase I 🦩🦩

ZeeJab:
This PR is the first phase of converting mobiledoc-kit into typescript. The motivation is to make the codebase more maintainable and understandable, and to allow us to make bigger changes with confidence.


tylr:
👍

PR #755, opened 2021-04-19
Add additional re-exports

RuslanZavacky:
Add exports to allow to build additional use-cases on top of mobiledoc-kit. Left comment https://github.com/bustle/mobiledoc-kit/pull/710#issuecomment-822427552 and realized that it seems to be possible to add more exports. 

This 3 exports should cover re-exports here in TryGhost - https://github.com/TryGhost/SDK/blob/main/packages/html-to-mobiledoc/lib/converter.js#L1 and this is something we also use in the similar fashion.

I am open for comments how to better approach this if this is not the way to re-export this internal bits.

### DIFF
--- a/src/js/index.ts
+++ b/src/js/index.ts
@@ -5,6 +5,8 @@ import Range from './utils/cursor/range'
 import Position from './utils/cursor/position'
 import Error from './utils/mobiledoc-error'
 import VERSION from './version'
-import { MOBILEDOC_VERSION } from './renderers/mobiledoc'
+import mobiledocRenderers, { MOBILEDOC_VERSION } from './renderers/mobiledoc'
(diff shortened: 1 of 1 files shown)


ZeeJab:
Hi! These are not part of the public API, they’re meant for internal use which is why they weren’t exported. But if you need them it seems fine. I can do a release later. 

We can also work on making them part of the public API.

RuslanZavacky:
They help in a way when we want to do Mobiledoc -> HTML and then HTML -> Mobiledoc translation. If they can be part of public API - that would great help (or analogue API, that can be public).

Thank you!

ZeeJab:
Everything you need should be in the latest release now @RuslanZavacky 👍

PR #764, opened 2022-08-11
Fix emoji input on iOS

JBYoshi:
Many emoji char codes are bigger than String.fromCharCode() supports. String.fromCodePoint() has a larger range that allows emojis.

Fixes TryGhost/Ghost#11541.


gpoitch:
Thank you. Updated the title as they work correctly on desktop. 

Tests are passing. The failure is due to account secrets not working in forks that we need to figure out.

PR #774, opened 2022-08-31
Add css in "exports" section of package.json

romgere:
Declare `mobiledoc.css`as an export of mobiledoc-kit library to ease import of css in ESM environment.


gpoitch:
Thank you

PR #784, opened 2022-09-13
Fix broken documentation links in README.md

stfnio:
Apparently, these links weren't updated when the codebase switched to TS.


stfnio:
@gpoitch Could you please take a look?

PR #789, opened 2023-09-18
Split shortcuts by platform and name them

lukemelia:
This is a rebase and modernization of @hpohlmeyer's #591.

Our use case is that we need to unregister the default toggleLink commands to replace them with our own more sophisticated link editing UI.


lukemelia:
CI failures are related to sauce labs credentials

PR #790, opened 2023-09-18
Export MobiledocParser and utils functions to support extensibility

lukemelia:
Our use case for these additional exports is as follows:

MobiledocParser: we use it in a "cleaning" service that processes mobiledoc before saving to remove cards that have been added but not configured and so are effectively empty and should be removed.

parsePostFromHTML: we use it to let users convert older HTML content to mobiledoc, and in our tests to allow succinct creation of mobiledoc from HTML 

parsePostFromText: we use it to convert text from a CSV import into mobiledoc for further editing in our UI 

detect: we use it to find link markups in a mobiledoc to facilitate selection expansion when replacing a link

### DIFF
--- a/src/js/index.ts
+++ b/src/js/index.ts
@@ -8,6 +8,9 @@ import Error from './utils/mobiledoc-error'
 import Renderer, { MOBILEDOC_VERSION } from './renderers/mobiledoc'
 import DOMParser from './parsers/dom'
 import PostNodeBuilder from './models/post-node-builder'
+import MobiledocParser from './parsers/mobiledoc'
+import { parsePostFromHTML, parsePostFromText } from './utils/parse-utils'
+import { detect } from './utils/array-utils'
 
 export {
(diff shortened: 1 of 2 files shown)


lukemelia:
Looks like the build failure here is related to SauceLabs credentials.

gpoitch:
Use cases sound good.

I believe you can do these with public api:

MobiledocParser: 
```js
const post = new Editor({ mobiledoc: { version: '0.3.2', atoms: [], cards: [], markups: [['strong']], sections: [[1, 'p', [[0, [], 0, 'Foo '], [0, [0], 1, 'bar'],],],],},}).loadPost()
```
parsePostFromHTML: 
```js
const post = new Editor({ html: '<p>Foo <b>bar</b></p>' }).loadPost()
```

parsePostFromText: could add a `text` option and perform like above.

detect: seems like a pretty basic utility to just copy

